PR #144, opened 2015-03-18
Update to Rails 4.2

tsujigiri:
I updated Rails to 4.2. I also updated all the other gems. All I can say is that the tests are green. If you are not confident this is enough, do thorough manual testing or... **write more tests!** :P

**TODO**: Add `secret_key_base` file on server.


app/controllers/fitbit_profiles_controller.rb
```diff
@@ -22,7 +22,9 @@ def show
     
     #grab activity measures for graphs
     if @fitbit_profile.activities
-      @activity = FitbitActivity.find_all_by_fitbit_profile_id(@fitbit_profile.id,:order => "date_logged")
+      @activity = FitbitActivity
+        .where(fitbit_profile_id: @fitbit_profile.id)
+        .order(:date_logged)
```
tsujigiri: That's so oldschool...

tsujigiri:
All the Rubocop preferences are configurable, btw.. See [bbatsov/rubocop](https://github.com/bbatsov/rubocop).

app/controllers/genotypes_controller.rb
```diff
@@ -47,7 +47,7 @@ def show
 
   def feed
     # for rss-feeds
-    @genotypes = Genotype.all(:order => "created_at DESC", :limit => 20)
+    @genotypes = Genotype.order('created_at DESC').limit(20)
```
tsujigiri: I changed that check in the rubocop config.

app/controllers/news_controller.rb
```diff
@@ -2,14 +2,14 @@ class NewsController < ApplicationController
...
     @title = "News"
-    @new_genotypes = Genotype.all(:order => "created_at DESC", :limit => 20)
-    @new_users = User.all(:order => "created_at DESC", :limit => 20)
-    @new_phenotypes = Phenotype.all(:order => "created_at DESC", :limit => 20)
-    @new_phenotype_comments = PhenotypeComment.all(:order => "created_at DESC", :limit => 20)
-    @new_snp_comments = SnpComment.all(:order => "created_at DESC", :limit => 20)
+    @new_genotypes = Genotype.order('created_at DESC').limit(20)
+    @new_users = User.order('created_at DESC').limit(20)
+    @new_phenotypes = Phenotype.order('created_at DESC').limit(20)
+    @new_phenotype_comments = PhenotypeComment.order('created_at DESC').limit(20)
```
tsujigiri: This is also up for debate. Some prefer 80, others 100.

philippbayer:
:D

Thank you very much for all of this work, it looks like a ton! Will try it out and see if I can find anything (probably not), and then deploy, cheers!

philippbayer:
So I assume I have to run

    rake secret > secret_key_base

once, right? This will probably break on deployment

tsujigiri:
Yes. We will have to put this file into `/srv/www/snpr/shared/config` and create a link via capistrano as we already do with the `secret_token` file.

tsujigiri:
Or we make use of the secrets.yml for this: http://guides.rubyonrails.org/security.html#session-storage

PR #150, opened 2015-04-05
Remove calls to `.all` with arguments

tsujigiri:
These are broken since the Rails 4 update.


app/controllers/picture_phenotypes_controller.rb
```diff
@@ -96,14 +96,16 @@ def show
     #@phenotypes = Phenotype.where(:user_id => current_user.id).all
     #@title = "Phenotypes"
     @phenotype = PicturePhenotype.find(params[:id]) || not_found
-    @comments = PicturePhenotypeComment.where(:picture_phenotype_id => params[:id]).all(:order => "created_at ASC")
+    @comments = PicturePhenotypeComment
+      .where(:picture_phenotype_id => params[:id])
+      .order(:created_at)
```
tsujigiri: Nope.

PR #155, opened 2015-04-07
Make ActiveJob and hence ActionMailer use Sidekiq

tsujigiri:
The default for ActiveJob is to run everything inline.


philippbayer:
This seems to break tests:

     undefined method `subject' for nil:NilClass

PR #172, opened 2015-05-14
Put workers accessing user_snps on same queue

### LINKED ISSUE #171: Only allow one job to access user_snps at a time
Currently I see two long-running jobs that access user_snps: Parsing and DeleteGenotype. Both started "about 5 hours ago". I guess it would make sense to find out if there are more and put them all on one queue, limited to one job at a time.

tsujigiri:
Closes #171

### DIFF
--- a/app/workers/parsing.rb
+++ b/app/workers/parsing.rb
@@ -1,6 +1,6 @@
 class Parsing
   include Sidekiq::Worker
-  sidekiq_options :queue => :parse, :retry => 5, :unique => true
+  sidekiq_options queue: :user_snps, retry: 5, unique: true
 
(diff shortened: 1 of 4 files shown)


app/workers/parsing.rb
```diff
@@ -108,7 +107,9 @@ def insert_into_user_snps
           #{temp_table_name}.local_genotype,
           #{genotype.id} as genotype_id
         from #{temp_table_name}
-        where #{temp_table_name}.snp_name not in (#{known_user_snps.to_sql})
+        where #{temp_table_name}.snp_name not in (
+          select snp_name from user_snps where genotype_id = #{genotype.id}
+        )
```
tsujigiri: I made this more explicit, so it's more clear that all the work is done by the database.

PR #175, opened 2015-05-17
Partition user_snps

### LINKED ISSUE #173: Partition the user_snps table
The user_snps table is approaching 1 billion rows, which leads to importing things into it taking several hours. Recently it fails a lot due to memory allocation errors, that may or may not be related to data corruption.

But despair not! Rescue is nigh!

What we can do, and what I refrained from so far, is [partitioning](http://www.postgresql.org/docs/9.2/interactive/ddl-partitioning.html) the table. It actually is the obvious thing to do, but Postgres doesn't make it as simple as e.g. MySQL does. But I guess we can't procrastinate this forever. ;)

My idea is, that we make a partition per genotype. This way we save the largest and currently almost always failing step of shoveling the data from the temporary table into the actual table. Instead we basically keep the temporary tables as partitions. We then also don't need the genotype_id column anymore. Parsing would become a matter of seconds instead of hours. The same goes for deleting a genotype's user-SNPs, we simply drop the table and they are gone.

What we need to do beforehand is figure out what information we actually need from that table and how to retrieve it when the table is partititoned. E.g. we need the number of user-SNPs per SNP which means one row from each table. We will have to dig into that. We could, for example, keep track of that in the snps table.

It would also be great if we can find a strategy that let's us first change the Parsing worker, so it already stores the new data to the partitions while we migrate the old data. It might even be a better idea to just parse all the files again instead of dealing with the humongous user_snps table.

I will start with the Parsing worker and try to figure out how to make it do the partitioning.

tsujigiri:
:construction: WORK IN PROGRESS :construction:
DO NOT MERGE, YET!

Closes #173


tsujigiri:
To get the number of `UserSnp`s per `Snp`, we can use the counter cache, we already have in place.
Do we need these as a list somewhere, too?

gedankenstuecke:
On the SNP show view there is a list of all UserSnps per Snp, right?

tsujigiri:
Indeed, there is. We should definitely test how this performs now before we deploy. I'll clean that up and test it first.

tsujigiri:
So, querying `user_snps` by `snp_name` has a time-complexity of `log(n) * m` with this partitioning scheme, where `n` is the number of rows per partition, which is roughly a million in our case, and `m` is the number of genotypes. So, if we assume ~1000 genotypes, this results in ~13815, where querying this data from a single table has a complexity of `log(n*m)` which is ~20. This doesn't really tell us what this means in terms of milliseconds, though. Either we drop this list for now, move on and figure something out later (or we don't), or we come up with something else.

gedankenstuecke:
What I initially forgot: we also need this access for the API, because there we also allow calls for getting all UserSnps at a given Snp. I assume that it's a rather central element of the API, so we should figure out a way to keep it in I'd say, but maybe I'm mistaken.

tsujigiri:
Alternatively, we could partition by `snp_name`, which would result in a million partitions with a thousand rows each. I suspect, querying this by genotype_id is not really an option. Maybe we can find a different database, that works better for this.

gedankenstuecke:
Let's try a graph database :p

tsujigiri:
At least you didn't say MongoDB...

gedankenstuecke:
Well, we are having a humongous amount of data points after all… :runner: :racehorse: 

But seriously, do you have a suggestion? :smile:

tsujigiri:
DynamoDB looks like it can do what we need, as long as we don't need to pay for it... ;)

tsujigiri:
But yeah, I'm investigating.

tsujigiri:
Do you have any experience with graph databases? I'm trying to figure out whether this is an actual option or not.

gedankenstuecke:
I have zero experience on it, though I heard a talk on them being used for genotype-phenotype-stuff last week. Which sounded very interesting, but their scale was a looooot smaller, as they only wanted to crosslink associations, not the actual genotype variations of 1000s of users.

gedankenstuecke:
Example

I did play around a bit with Neo4j today and in principle I can see the appeal of using a graph db for our data. You could easily manage to convert the current schema into a graph as shown here I think.

With () being nodes and [] the edges w/ their descriptions:

(User) <-[uploaded_by]- (genotype_file) <-[found_in with_genotype]- (SNP)

By doing so we could store all the information that's currently put into the UserSnps as edges between Snps & GenotypeFiles. But I've no clue whether this would be more efficient. Instead of joining some tables we now would have to explore the graph neighborhood I guess. :smile_cat:

tsujigiri:
In the case of user_snps, we don't even join it anymore due to its size. The way I imagine it could work with a graph database is that we would define edges both ways between Snp and Genotype. This is probably a lot of redundant information, but so is an index in a relational databse. Probably worth trying with actual data.

gedankenstuecke:
I'm not sure what you mean by having edges for both ways. You can have bidirectional edges, e.g. http://graphaware.com/neo4j/2013/10/11/neo4j-bidirectional-relationships.html

gedankenstuecke:
Or to put it more correctly: you can only do directed edges, but you can query ignoring the direction, so in the end it's ± bidirectional.

tsujigiri:
Ah, I see. Cool. :)

gedankenstuecke:
yeah, performance/speed-wise I'm not sure how it would compare to our current setup for now. But I agree that it probably still will take more memory, as the attribute-descriptors of edges/nodes are kept for each edge/node and not only in the column-name.

tsujigiri:
So, I played around with Neo4j and importing stuff is very slow. I didn't get transactions working with the Ruby library. The documentation isn't very well documented and it looks like with every major version they just change about everything.

I had a different idea though: We could add array/hstore columns to snps/genotypes, with the snps' array storing a list of associated genotypes and the genotypes' hstore storing a hash of `snp_name => local_genotype` pairs. This way it is nicely *partitioned*, we don't need any indexing (I think). I started implementing this [here](https://github.com/gedankenstuecke/snpr/compare/master...tsujigiri:use-array-and-hstore?w=1). So far the parser works and is fast. I'm not around for the next two weeks. Feel free to move on with this.

PR #177, opened 2015-06-11
Added Beacon implementation

gedankenstuecke:
Beacon's are a pretty simple API implementation designed by the Global Alliance 4 Genomics & Health. Basically you can ask for a given chromosome + position on whether a given allele is found in a database. So for example you can ask "for chromosome 1, position 1234, do you have any individual with an "A" in your DB?". There are only 3 plain text answers: "YES","NO","NONE" (the latter for cases where things didn't work out, incorrect API calls etc). A bit more details can be found here: http://dnastack.com/ga4gh/bob/subscribe.html

In any case, this adds support for this type of API query and it even has test coverage. If you start your dev version you should be able to test it using localhost:3000/beacon/rest/responses?pos=72017&chrom=1&allele=A etc.

### DIFF
--- a/test/functional/beacon_controller_test.rb
+++ b/test/functional/beacon_controller_test.rb
@@ -1,0 +1,27 @@
+# encoding: utf-8
+require_relative '../test_helper'
+
+class BeaconControllerTest < ActionController::TestCase
+  context "Beacon" do
+    setup do
+      activate_authlogic
+      Sidekiq::Client.stubs(:enqueue)
+      @user = FactoryGirl.create(:user)
+      @snp = FactoryGirl.create(:snp)
+      @snp.allele_frequency["A"] = 2
+      @controller.send(:reset_session)
+    end
+    should "be YES" do
+      get(:responses, {pos: @snp.position,chrom: @snp.chromosome,allele: "A")
+      assert_equal("YES",response.body)
+    end
+    should "be NO" do
+      get(:responses, {pos: @snp.position,chrom: @snp.chromosome,allele: "C")
+        assert_equal("NO",response.body)
+    end
+    should "be NONE" do
+      get(:responses, {pos: @snp.position,chrom: @snp.chromosome)
+        assert_equal("NONE",response.body)
+    end
+  end
+end
--- a/test/functional/beacon_controller_test.rb
+++ b/test/functional/beacon_controller_test.rb
@@ -9,6 +9,7 @@ class BeaconControllerTest < ActionController::TestCase
       @user = FactoryGirl.create(:user)
       @snp = FactoryGirl.create(:snp)
       @snp.allele_frequency["A"] = 2
+      @snp.save
       @controller.send(:reset_session)
     end
     should "be YES" do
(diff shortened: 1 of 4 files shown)


test/functional/beacon_controller_test.rb
```diff
@@ -0,0 +1,28 @@
...
+    should 'be NO' do
+      get(:responses, {pos: @snp.position, chrom: @snp.chromosome, allele: "C"})
+        assert_equal('NO',response.body)
+    end
+    should 'be NONE' do
+      get(:responses, {pos: @snp.position, chrom: @snp.chromosome})
+        assert_equal('NONE',response.body)
+    end
+  end
+end
```
tsujigiri: Prefer RSpec tests over Test::Unit ones. *nag nag nag*

tsujigiri:
Looks OK. If you could try satisfy the Automatic Complainer™, that would be great. ;)

test/functional/beacon_controller_test.rb
tsujigiri: But then, better than no test at all. ;)

gedankenstuecke: What's the reason to prefer RSpec? (as i said: I have no idea what I'm doing, adapting this was the best I could come up with so far ;-))

gedankenstuecke:
Yeah, I tried to fix some of them :)

tsujigiri:
OK, I can try to be less pedantic about these things, but in my experience it always pays off in the long run to put some effort into it. Ask me about it over a beer. ;)

test/functional/beacon_controller_test.rb
tsujigiri: It's nicer to work with and has become sort of a de facto standard. We also agreed on it a while ago. But, as I said, any test is better than no test. I can move it over to RSpec, so you can see how it would translate. It's not all that important, though. We still have the user-SNP issue. I'm working on this one at the moment.

gedankenstuecke:
No, it's completely fine. I'm happy for any code critique! Btw: Did you see the weird workaround I did for the migration? With those lines present it just failed to db:migrate thanks to a missing gem. Is the way I solved it the way to go in those cases?

test/functional/beacon_controller_test.rb
gedankenstuecke: Great, I must admit I forgot that we had agreed to RSpec and seeing both tests I ran with the one that felt easier. :sheep:

tsujigiri:
Didn't we get rid of ActiveAdmin?

tsujigiri:
Ah, it's an old migration. Why would you want to change it? Migrations are usually a one time thing. It won't run on the server either, as it already did.

gedankenstuecke:
right, but if you set up a new machine (as I just did), you can’t just migrate up with the old migration because it just crashes. thought might be nice for people who want to set up a dev version on their end.

tsujigiri:
That's what `db:setup` is for. Otherwise you would have to keep your whole code base backwards compatible.

-------- Original message --------

gedankenstuecke:
The more you know! https://media4.giphy.com/media/EldfH1VJdbrwY/200.gif <https://media4.giphy.com/media/EldfH1VJdbrwY/200.gif>

gedankenstuecke:
btw: what do you say, ready for merging into master and deploying?

tsujigiri:
Go ahead! :shipit:

PR #180, opened 2015-06-23
added instructions for getting started and running tests

audy:
I wanted to see if I could run the snpr tests. The instructions for running snpr weren't completely clear so I took notes along the way and added them to the readme.


audy:
Also, just a minor suggestion: It would remove a step and thus making things easier if `config/database.yml.example` and `config/app_config.yml.example` were just config files with sane defaults for development.

tsujigiri:
This is great! We definitely have neglected the README recently. We should get rid of **all** the files that one has to create before starting the app and replace them with sane defaults for development.

tsujigiri:
@gedankenstuecke @philippbayer So, are you guys OK with merging this?

gedankenstuecke:
Sure, I'm all for it. :-)

PR #182, opened 2015-06-27
Move list of SNPedia links into partial

tsujigiri:


app/models/snpedia_paper.rb
```diff
@@ -9,4 +9,13 @@ class SnpediaPaper < ActiveRecord::Base
...
   end
+
+  def matches_local_genotype?(other_local_genotype)
+    local_genotype == other_local_genotype.split('').sort.join
+  end
+
+  def local_genotype
+    url =~ /\((.*);(.*)\)$/
+    $1 + $2
+  end
```
tsujigiri: Is `local_genotype` an appropriate name here?

gedankenstuecke: Depends a bit on the context, what exactly are you going for?

tsujigiri: It's extracting the nucleotide pair from the SNPedia URL. The first method compares this with the `local_gentype` from the `UserSnp`, sorting the letters alphabetically, because they seem to be sorted in SNPedia as well.

tsujigiri: SNPedia links are displayed bold if the nucleotide pair matches a SNP in your genotype. I moved this logic here. It is used [here](https://github.com/gedankenstuecke/snpr/pull/182/files#diff-8c0ff2a01904903c17a5dd6e62a2cc1eR3).

app/views/snps/_snpedia_papers.html.erb
```diff
@@ -0,0 +1,27 @@
...
+          </td>
+          <td><b><%= p.summary.capitalize %></b></td>
+        </tr>
+      <% end %>
+    </table>
+  <% else %>
+    There are no SNPedia-links yet. Please come back later.
+  <% end %>
+  <div class="small">
+    Note: For some SNPs, testing providers detect the genotype from the opposite strand of DNA, so that the genotypes listed at openSNP don't appear here. In those cases, please replace "A" by "T" or "G" by "C" (and vice-versa).
```
tsujigiri: What does this mean, btw.? Users should substitute A with T or G with C in their heads? Couldn't this be incorporated into the matching, so that links are emphasized that match, but were detected "from the opposite strand of DNA"?

tsujigiri: Do users even understand what the emphasis means? It isn't eplained anywhere, is it?

gedankenstuecke: Yes, this means the users have to do the substitution in their head. You could try to match for the opposite strand in some instances. But there's the problem where the variants are C & G or A and T and in those cases it's impossible to decide whether to highlight the flipped strand or the given one. Does that make sense?

gedankenstuecke: Oh, and you're right. It's not explained anywhere. :D

tsujigiri: Not sure I understand. In the case of SNPedia it seems to me like the variant is always ordered alphabetically anyway (does this even make sense, biologically?). If you flip and re-order "CG", it will still be "CG".

gedankenstuecke: Ah, okay. So we're talking about actually a biological strand flip here. Not the ordering of the two alleles in a genotype. 

Whether it's "GC" or "CG" is indeed irrelevant for this and we already should check for this at this point w/ SNPedia iirc.

But: You know that C <-> G and T <-> A pair in the double stranded DNA. So when 23andMe or other companies read out your SNPs it actually matters which of the two strands they are looking at. In the easy case you can have 3 genotypes at a given SNP, let's say AA, AG or GG. If you now have a user who is TC, TT or CC you can directly infer that the company reported the opposite strand compared to SNPedia. 

With me so far? Okay, now imagine the genotypes are CC, CG or GG. In that case you have no way of figuring out whether the genotype you are looking is, is from the correct strand. 

Does this make more sense? :-)

app/models/snpedia_paper.rb
tsujigiri: So, what about the name? Do you have a better suggestion or is this fine? Irrespective of the use case, what would you call the two letters in a SNPedia URL? On their pages it's just "genotype". Is this better than "local genotype"?

gedankenstuecke: > So, what about the name? Do you have a better suggestion or is this fine? Irrespective of the use case, what would you call the two letters in a SNPedia URL? On their pages it's just "genotype". Is this better than "local genotype"?

I think the problem is that we’re also using “Genotype” for the files, so we would maybe overload the term when using it there as well? I think “genotypes/variations at SNP” would be the best if not nicest fitting name. But “local genotype” should also be fine.

Cheers,
Bastian

app/views/snps/_snpedia_papers.html.erb
tsujigiri: I think so. The one thing I apparently didn't get yet, though, is why there are always two *letters*. Whenever I read up on SNPs, it says SNPs are *single* alleles that differ between individuals. Then there are the variations, each consisting of two instead of one letter.

do_you_even_biology.jpg

app/models/snpedia_paper.rb
tsujigiri: Hm... how about `snp_variation`, then?

gedankenstuecke: > On 08 Jul 2015, at 08:32, Helge Rausch <notifications@github.com> wrote:
> 
> Hm... how about snp_variation, then?

Works for me :-)

gedankenstuecke:
btw. should this be merged now, or do you want to/shall we change the names first? :)

tsujigiri:
I changed the name of `SnpediaPaper#local_genotype` to `SnpediaPaper#snp_variation`. Unless there is anything else you think could be renamed, this is good to go.

gedankenstuecke:
here we go :dog:

PR #185, opened 2015-06-29
removed vegas from Gemfile for #184

### LINKED ISSUE #184: Remove Vegas from Gemfile?
This gem is included in the `Gemfile` but does not appear to be used in the app (the only hits for "Vegas" in the source are `Gemfile` and `Gemfile.lock`). Also, the gem has not been updated since 2012. All specs pass without it.

audy:
Closes #184

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -350,8 +350,6 @@ GEM
     unf_ext (0.0.7.1)
     uuidtools (2.1.5)
     vcr (2.9.3)
-    vegas (0.1.11)
-      rack (>= 1.0.0)
     webmock (1.21.0)
       addressable (>= 2.3.6)
       crack (>= 0.3.2)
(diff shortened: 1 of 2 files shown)


Gemfile.lock
```diff
@@ -422,11 +420,7 @@ DEPENDENCIES
...
   uuidtools
   vcr
-  vegas
   webmock
   whenever
   will_paginate
   yui-compressor
-
-BUNDLED WITH
-   1.10.2
```
tsujigiri: Update bundler and try again. :stuck_out_tongue:

audy: / (╯°□°）╯︵ rbenv

tsujigiri: It's just bundler. It got introduced with 1.10 and now, I think, you get a warning or even an error if your bundler version is older than this.

PR #189, opened 2015-07-19
Store "user-SNPs" in HSTORE columns

tsujigiri:
:warning: Do not merge, yet! See **Migration**. :warning: 

Due to it's size, the `user_snps` table got pretty much unmanagable. Parsing a genotyping and inserting the resulting user-SNPs into the table took several hours. To speed up the import, I moved the information it contains into HSTORE columns on `snps` and `genotypes`. Import is now a matter of seconds, while querying the data should remain about as fast as before or even faster.

* The `snps` table gets a `genotypes` column. Each row contains an HSTORE of all associated genotype_ids and their local genotype at the respective SNP (E.g. "123" -> "GC").
* The `genotypes` table gets a `snps` column. Each row contains an HSTORE of all associated SNPs by name and the genotype's local genotype for the respective SNP (E.g. "rs123" -> "GC").

This means we store the local genotype twice for each Genotype <-> Snp association. This serves performance when querying the local genotype from either side.

#### On class level

`Snp` and `Genotype` each have a new scope `with_local_genotype_for`, taking one argument. In case of `Snp` the argument can either be an instance of `Genotype` or a `Genotype#id`. In case of `Genotype` it can either be an instance of `Snp` or a `Snp#name`. It adds a `local_genotype` attribute to the instances.

```ruby
Snp.with_local_genotype_for(genotype).each { |snp| snp.local_genotype }
Genotype.with_local_genotype_for(snp).each { |genotype| genotype.local_genotype }
```

#### On instance level

Querying associated `Snp`s for an instance of `Genotype` or vice versa does not work anymore through the usual ActiveRecord associations. Instead there are instance methods on each side, taking care of the querying.

```ruby
Snp.find_by(name: 'rs123').genotypes
Genotype.find_by(id: 123).snps
```
### Migration

In order to migrate to the new schema, two things need to happen. `Genotype` and `Snp` have a default scope, excluding the new HSTORE columns from all queries they are not needed for. This prevents the HSTORE getting serialized into a Ruby Hash every time an instance is created. I will create a separate pull request for that. Then, the new version needs to be copied onto the server, but not yet switched to. We need to run themigration and re-import all `Genotype`s first. After this is done, we can switch over. If I didn't miss anything, no one will notice the disruption in the force. ;)


app/controllers/das_controller.rb
```diff
@@ -1,123 +1,122 @@
...
+  def startpoint
+    @user = User.find_by_id(params[:id])
+    response.headers['X-DAS-Version'] = 'DAS/1.53E'
+    response.headers['X-DAS-Status'] = '200'
+    response.headers['X-DAS-Capabilities'] = 'features/1.1; sources/1.0'
+    response.headers['X-DAS-Server'] = request.env['SERVER_SOFTWARE'].split(' ')[0]
+
+    render template: 'das/startpoint.xhtml.erb', layout: false
+  end
+end
```
tsujigiri: I didn't want to go down this rabbit hole. I just fixed the indentation and made it work with the new schema.

gedankenstuecke:
Wow, well done!

app/controllers/das_controller.rb
gedankenstuecke: totally understandable! I wouldn't want to go there too soon either :P

philippbayer:
Thank you very much for this extremely detailed write-up!

philippbayer:
Hi,

I just tried to get this branch to run on my machine, I had to `sudo apt-get install postgresql-contrib` for the HSTORE extension, else you'll get

    ActiveRecord::StatementInvalid: PG::UndefinedFile: 
    ERROR:  could not open extension control file "/usr/share/pgsql/extension/hstore.control": No such file or directory

We may run into issues with this on the server as the PostgreSQL installation is not the newest, so the extension may not be available. In that case it may be easier and way faster to drop the UserSNP table and then to upgrade Postgresql, but that depends on how fast the new system loads userSNPs.

Also, is it intended that the queue "user_snps" gets the parsing job? It's not a default queue in the serverscript, for example.

Just tried to parse "testdatensatz3_23andme.txt" and while that correctly created 43 SNPs and User.first.snps.length equals 43, as well. I think this is good to be deployed?

tsujigiri:
Upgrading Postgres is generally a good idea, I think the hstore extension isn't that new, though. Apart from that it should be good to go.

philippbayer:
I'm a bit scared I run into the same thing as last time, when we set the database to "read only" while postgresql upgraded the data, and it took forever.

Can I merge both pull requests?

tsujigiri:
#190 should be merged and deployed first. Then this can be merged and copied to the server, but not switched to, yet. Run the migrations and then switch afterwards. The migration shouldn't take too long either.

philippbayer:
I've merged #190, I have to submit my PhD thesis tomorrow so I might not get around to doing it this weekend :P

audy:
:shipit: @philippbayer !

philippbayer:
I'd love to ship it if it weren't for tiny typos in these 50,000 words

tsujigiri:
I can do it. The deployment I mean... :P

philippbayer:
That would be wonderful! 

I think we should write this up for the blog once we have some parsing benchmarks, something like "Using Hstore to speed up parsing and retrieval of ~millions of records by X" or so. Any volunteers?

and yeaaaah if you could just go ahead and finish this thesis for me, that would be great :necktie:

gedankenstuecke:
Awesome, thanks @tsujigiri! :+1: 
And best of luck with submitting, @philippbayer!

philippbayer:
I've submitted. Has #190 been deployed?

tsujigiri:
It's deployed.

philippbayer:
:+1:

tsujigiri:
I'll be running the migrations on this one now... :mask:

tsujigiri:
Starting the parsing...

tsujigiri:
`Parsing#update_snps` already takes an hour for the first Genotype and it's still running. :/

tsujigiri:
So, I ran it over night and the parsing times are varying a lot. From what I've seen so far between 20 and 60 minutes. IIRC it took 2 minutes on my machine with a full snps table and on an SSD.
There is still a few errors, but overall that's better than before, I guess.

philippbayer:
Does it even out over time? I.e., did the first genotypes take longer than the later ones? It's definitely better than before!

gedankenstuecke:
Sweet! That means in the worst case we have to spend 80 days of parsing? :D

tsujigiri:
I'm not sure what makes the difference in parse time. So far I don't see a pattern. Maybe it's just number of entries in the file. I'll investigate.

tsujigiri:
A more thorough look showed that the parse time pretty linearily reflects the number of rows. Most have about 500k 600k and take about 20 minutes.

tsujigiri:
:tada:

gedankenstuecke:
Very nice! Well done! :smile: :+1: :fireworks:

philippbayer:
20 minutes is an extreme improvement!!

Great job!

~2100 genotypings * 20 min = 42,000 min, 700 hours, 29 days. Do we have to run only one job at a time? Each COMMIT seems to take about 15% RAM

tsujigiri:
I assume the bottleneck to be the disk, so I don't think we can speed up things by parsing in parallel. Faster disks would help, though. ;)

gedankenstuecke:
which reminds me, to remind you to read your emails re: funding ;)

tsujigiri:
I saw it, I will respond. tl;dr: I like it. ;)

gedankenstuecke:
sweet! :dancer: alternative ideas on how to "earn" money are also appreciated  :wink:

tsujigiri:
The disks are really slow even with tasks as mundane as copying the bundle directory into the new release's directory, to save some time skipping the `bundle install`. In the end it takes twice as long because it doesn't only have to be written but also read... We should eventually try that switching to a new server, hoping it has faster disks...

tsujigiri:
I'll close this *for now*...

PR #190, opened 2015-07-19
Add default scopes, excluding HSTORE columns

tsujigiri:
In preparation for #189, exclude the coming HSTORE columns, so we can migrate in the background, without disrupting service.

### DIFF
--- a/app/workers/preparsing.rb
+++ b/app/workers/preparsing.rb
@@ -82,7 +82,7 @@ class Preparsing
     logger.info "Checking whether genotyping is duplicate"
     md5 = Digest::MD5.file("#{Rails.root}/public/data/#{genotype.fs_filename}").to_s
     file_is_duplicate = false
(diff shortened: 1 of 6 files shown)


app/workers/preparsing.rb
```diff
@@ -82,7 +82,7 @@ def perform(genotype_id)
     logger.info "Checking whether genotyping is duplicate"
     md5 = Digest::MD5.file("#{Rails.root}/public/data/#{genotype.fs_filename}").to_s
     file_is_duplicate = false
-    if Genotype.where(md5sum: md5).where('id != ?', genotype.id).count > 0
+    if Genotype.unscoped.where(md5sum: md5).where('id != ?', genotype.id).count > 0
```
tsujigiri: This is the only case the above workaround didn't work for, so I had to add the `unscoped`. I have no clue why.

PR #191, opened 2015-09-26
Add .rubocop_todo.yml

tsujigiri:
In order for rubocop complaints to have any meaning I had rubocop
generate this file. We can now run `rubocop` locally and it should be
green unless we introduced new issues. For good karma, remove
lines from it, run rubocop and fix the issues it complains about. :)

Read this for more information:
https://github.com/bbatsov/rubocop#automatically-generated-configuration


tsujigiri:
Does anyone mind if I merge this? Being able to just call `rubocop` locally and see what issues you introduced instead of all the issues would make this much more useful.

philippbayer:
+1

gedankenstuecke:
+1

philippbayer:
FYI:

1) Had to run `gem install rubocop` after `bundle install` as the executable wasn't there

2) Had to run `rubocop -R` to get the Rails-relevant rules, else it wouldn't complain :P

tsujigiri:
It's not supposed to complain about things in the todo file. I missed the -R flag, though. I will regenerate the todo file with the flag.

tsujigiri:
I guess we should add some info on that to the "Contribute" section.

PR #192, opened 2015-09-26
Make tests fail if test coverage drops

tsujigiri:
If I did this correctly, the tests should always fail, if the test-coverage drops. If it increases, `coverage/.last_run.json` should be updated automatically. Don't forget to commit changes to it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,8 @@ api_key.txt
 config/app_config.yml
 config/database.yml
 config/newrelic.yml
-coverage
+coverage/*
+!coverage/.last_run.json
 db/*.sqlite3
 development.log
 dump.rdb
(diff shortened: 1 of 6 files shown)


tsujigiri:
The Ruby 2.0 build failed on TravisCI due to a lower test coverage. Since we don't really need it I just removed it. What do you think about removing the 2.1 build as well?

tsujigiri:
Now they all fail... hm...

tsujigiri:
In Ruby 2.2 the coverage is slightly higher, so I set it to the value of the previous Ruby versions to make them all pass. I suppose it would be more practical to stick to one Ruby version, though. What do you think? Should I remove the other ones? Do you think forbidding coverage drops is a good idea?

gedankenstuecke:
Sounds good to me! Forces me writing tests ;)

tsujigiri:
I'll just merge this then.

PR #193, opened 2015-09-26
Use new TravisCI infrastructure

tsujigiri:
This lets us use TravisCI's new container-based infrastructure, which should speed up the builds significantly due to caching for bundler.

See: http://docs.travis-ci.com/user/migrating-from-legacy/


tsujigiri:
Since no one objected yet, I'll just assume it's fine for everyone, as it is a trivial change anyway. :)

tsujigiri:
First run with previously cached bundle reduced the build time from almost 4 minutes to just over 1. :clap:

PR #197, opened 2015-10-04
Moved factories into individual files in spec/factories

Calvyn82:
I haven't added any new factories, yet. I've only moved the existing factories over.


Calvyn82:
It's breaking tests right now. I'm going to fix that.

PR #198, opened 2015-10-04
Moved Factories and deleted redundancies

Calvyn82:
The tests should work fine now.


Calvyn82:
Do I need to go make houndci happy?

tsujigiri:
It is much appreciated for sake of consistency. :)

tsujigiri:
Looks great otherwise! Thanks a lot!

Calvyn82:
There a way to check hound locally before opening a pull request?

tsujigiri:
If you install the `rubocop` gem, you should be able to run `rubocop spec/factories` locally to see if it is satisfied. And you don't need to open a pull-request if you want to update it. You can just push to the same branch and the PR will be updated automatically. Force-pushing works as well.

Calvyn82:
I installed and ran it against that last pull request. I'll do it on all of my future commits.

Sorry about the pull request thing. If it isn't obvious, I'm still pretty new to all of this. Only been coding at all for about year, and this is only the second time I've submitted to a project I'm not a contributor on.

tsujigiri:
Don't worry at all! I totally agree that it's not obvious.

Issue #202, opened 2015-10-09
clean up papers


gedankenstuecke:
I noticed that we have tons of duplicates amongst the papers in our DB. E.g. https://opensnp.org/snps/rs9939609#plos

Here basically all papers there are listed at least twice. We should maybe go through each paper category once and check for duplicate names.

tsujigiri:
Do we maybe import papers that reference more than one SNP once for each of the SNPs?

gedankenstuecke:
I think originally we didn't, or at least then they should be only keyed to one SNP and not multiple ones at the same time. So I guess at some point the parser went wrong and started collecting papers multiple times for one SNP.

philippbayer:
The absurd thing is that they all have the same ID in the database:

``` ruby
Snp.find_by_name("rs9939609").plos_papers.each do |u|
   if u.first_author.include? "Yan Guo"
     puts u.inspect
   end
end; nil
```
says

```
#<PlosPaper id: 1284, first_author: "Yan Guo", title: "The Fat Mass and Obesity Associated Gene, FTO, Is ...", doi: "10.1371/journal.pone.0027312", pub_date: "2011-11-18 00:00:00", created_at: "2012-02-06 00:12:36", updated_at: "2014-10-30 16:08:59", reader: 3786>
#<PlosPaper id: 1284, first_author: "Yan Guo", title: "The Fat Mass and Obesity Associated Gene, FTO, Is ...", doi: "10.1371/journal.pone.0027312", pub_date: "2011-11-18 00:00:00", created_at: "2012-02-06 00:12:36", updated_at: "2014-10-30 16:08:59", reader: 3786>
#<PlosPaper id: 1284, first_author: "Yan Guo", title: "The Fat Mass and Obesity Associated Gene, FTO, Is ...", doi: "10.1371/journal.pone.0027312", pub_date: "2011-11-18 00:00:00", created_at: "2012-02-06 00:12:36", updated_at: "2014-10-30 16:08:59", reader: 3786>
and 8 more identical lines
```

These all have the same doi so I would have assumed that the worker's ```find_or_initialize_by_doi()``` would return these but that doesn't seem to be the case? ```find_by_doi()``` stills works and returns the first one of the above.

We could circumvent this problem by using ```validates :doi, uniqueness: true``` in all paper models, but that feels sloppy.

philippbayer:
Turns out the rake task doesn't work because you can't just fiddle with ActiveRecord::Relation.

[...]

PR #203, opened 2015-10-10
Makes sure that papers' SNPs are unique. Adds non-working rake task to make papers' SNPs unique.

philippbayer:
This should hopefully prevent #202 from happening again 

The Rake task doesn't work right now for some strange reason. It calls commit and everything but the paper remains "broken". I should drink more coffee.

(We can circumvent this issue by just storing all snps in a set instead of an array, but it's not really an array but a ActiveRecord::Associations::CollectionProxy)

### DIFF
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 55.81
+    "covered_percent": 51.41
   }
 }
--- a/coverage/.last_run.json
+++ b/coverage/.last_run.json
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 51.41
+    "covered_percent": 55.81
   }
 }
--- a/app/workers/mendeley_search.rb
+++ b/app/workers/mendeley_search.rb
@@ -74,7 +74,7 @@ class MendeleySearch
           logger.error("MendeleyPaper for #{snp.name} invalid.\n" <<
                        mendeley_paper.errors.full_messages.join(", "))
         else
-          mendeley_paper.snps << snp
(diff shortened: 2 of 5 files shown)


philippbayer:
Build fails because: Coverage has dropped by 4.40% since the last time 

I assume the 4% are added by the untested rake task? How do you even test rake tasks?

tsujigiri:
I'm getting the feeling that the coverage varies between machines. On your machine, simplecov seems to have increased the coverage, which made it fail on TravisCI.

coverage/.last_run.json
```diff
@@ -1,5 +1,5 @@
 {
   "result": {
-    "covered_percent": 51.41
+    "covered_percent": 55.81
```
tsujigiri: Just revert that change for now and you should be good. Are you using the same Ruby version as TravisCI does?

philippbayer:
Even with ```rvm use 2.2 ``` (after installing everything) I get the 55% coverage. Reverted it for now.

What I don't understand - the current .last_coverage file has 51%, my commit had 55% coverage. It complained that coverage dropped by 4%, but isn't it the other way around?

tsujigiri:
If the file says 55% and the tests say 51%, the coverage dropped by 4%, no?

I just changed the Ruby version in the .travis.yml from "2.2" to "2.2.0" to see if TravisCI would use the latest patch-level instead of "2.2.0-p0", but it didn't. It still used the same patch level. The tests failed though because of a coverage drop. Nothing else changed. I would say, let's drop this for now. It's not really useful if it isn't deterministic, is it?

philippbayer:
Yes, it does seem to be quite random!

tsujigiri:
See #204

PR #204, opened 2015-10-10
Revert "Make tests fail if test coverage drops"

tsujigiri:
This reverts commit 3142abe96dffec6cfc4a86fab4602c9a45458ef3.

Since the coverage metric doesn't seem to be consistent across different machines/Ruby versions/builds, let's revert this for now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,8 +3,7 @@ api_key.txt
 config/app_config.yml
 config/database.yml
 config/newrelic.yml
-coverage/*
-!coverage/.last_run.json
+coverage
 db/*.sqlite3
 development.log
 dump.rdb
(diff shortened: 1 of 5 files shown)


philippbayer:
It seems this issue is relevant:

https://github.com/colszowka/simplecov/issues/360

> I forgot that by default Travis configures bundler to install dependencies in vendor

So everything in vendor is then added to the test coverage? For RVM stuff is installed in ~/.rvm so it's of course not added

tsujigiri:
If this was the case, I would expect the difference to be bigger than 4%.

philippbayer:
That's entirely possible - let's just kick it out for now

PR #205, opened 2015-10-20
Move user_snps importing SQL into database

tsujigiri:
By upserting the user_snps instead of doing a sub-select, we should see some speedup in the Parsing worker. According to the internets, the performance penalty for using a for loop is insignificant compared to a plain insert.


philippbayer:
(Internets et al. 2015)

tsujigiri:
Correct citation is important! ;)

tsujigiri:
Should we switch on all the queues again and see what happens?

philippbayer:
Yes! That should work fine now, shouldn't it? I can make some graphs tomorrow to check

tsujigiri:
We should empty the mailers queue first. The Sidekiq Web UI is broken, though.

philippbayer:
This is from the apache log:

    Could not find rake-10.4.2 in any of the sources (Bundler::GemNotFound)

Looks like bundle install didn't work out correctly

tsujigiri:
I cannot look into it right now. If you have time, feel free to fix it.
I'm very curious to see if this change helps. :) *encourage*

philippbayer:
We're back now - Apache was still pointing to the old ruby 2.0.0 in /etc/apache2/apache2.conf

Sidekiq is now broken for different reasons :neckbeard:

philippbayer:
Turns out, a recent Sidekiq update changed some security-related things which broke our sidekiq, I implemented the workaround from here: https://github.com/mperham/sidekiq/issues/2459

However, now cap deploy doesn't work:

```
  * 2015-10-21 17:22:39 executing `deploy:migrate'
  * executing "cd /srv/www/snpr/releases/20151021092056 && bundle exec rake RAILS_ENV=production  db:migrate"
    servers: ["opensnp.org"]
    [opensnp.org:57329] executing command
*** [err :: opensnp.org:57329] pg_dump:
*** [err :: opensnp.org:57329] server version: 9.3.9; pg_dump version: 9.2.4
*** [err :: opensnp.org:57329] pg_dump: aborting because of server version mismatch
```

I'm not sure what this means - does the new database server run PG 9.3.9, and the webserver runs 9.2.4? If so, I think it's time for a system upgrade of the production server?

tsujigiri:
Looks like it. We should upgrade the system more frequently anyway. Be
aware that it might break things. ;)

philippbayer:
OK :) I already ran into some "fun problems", it's a rabbit hole of broken stuff today.

I'll also run a full release upgrade to Ubuntu 14-whatever's current for newest kernel version etc, is the entire database now on the database server?

tsujigiri:
Not sure it's really worth the effort since we will be moving the app to a new server anyway. Does upgrading postgresql-client work? Also, isn't the server already at the latest LTS release?

philippbayer:
```
$ cat /etc/lsb-release
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=10.04
DISTRIB_CODENAME=lucid
DISTRIB_DESCRIPTION="Ubuntu 10.04.4 LTS"
```
There's definitely some LTS versions after that ;)

The problem with updating postgresql-client from the current 9.2 is this:

```
The following packages have unmet dependencies:
  postgresql-client-9.3: Depends: libpq5 (>= 9.3.10) but 9.2.4-1.pgdg10.4+1 is to be installed
                         Depends: libssl1.0.0 (>= 1.0.0) but it is not installable
```

If I'm interpreting [this page](http://packages.ubuntu.com/search?keywords=libssl1.0.0) correctly then libssl1 is only available from Ubuntu 12 onwards

tsujigiri:
Oh, that's old... Maybe upgrading isn't such a bad idea. You could also
put the effort into making the app run in a Docker container, though, if
you have some time.
https://github.com/phusion/passenger-docker seems to be a good option.

philippbayer:
Welp, time to learn more about Docker ;) When are we moving to the new server? I'll upgrade the server once I ran another back-up of the data

tsujigiri:
I started with the Docker thing, but didn't really have time yet, so I
didn't get far. The plan is to move soon, maybe even in time to get rid
of the old server before the end of the month, so we can trash it and
save another month of costs for it.

philippbayer:
End of the month would be great! I tried to upgrade the whole distribution but 10.x is too old for a properly working ```do-release-upgrade```, the work-around would be to re-install the newest version, which is overkill if we're moving in 3 weeks anyway.

tsujigiri:
Alternatively we could just change the capistrano config to deploy to the new machine and take care of Docker later.

philippbayer:
That would be the easiest solution! Is the new machine set up?

tsujigiri:
Well, we would still need to set up everything else there as well, so it's basically the same effort as doing it in Docker. I'm already on the Docker thing. I just hope I can at least have it running in time and improve it later.

PR #207, opened 2015-10-28
Adds uniqueness constraint on paper-SNP connection. 

philippbayer:
Replaces rake task by SQL command.

See #202

### DIFF
--- a/app/models/snp_reference.rb
+++ b/app/models/snp_reference.rb
@@ -3,5 +3,5 @@ class SnpReference < ActiveRecord::Base
   belongs_to :snp
   belongs_to :paper, polymorphic: true
   #validates_presence_of :snp, :paper
-  validates_uniqueness_of :snp_id, scope: [:paper_id, :paper_type]
+  validates :snp_id, :uniqueness => { :scope => [:paper_id, :paper_type]}
 end
--- a/app/models/snp_reference.rb
+++ b/app/models/snp_reference.rb
@@ -3,5 +3,5 @@ class SnpReference < ActiveRecord::Base
   belongs_to :snp
   belongs_to :paper, polymorphic: true
   #validates_presence_of :snp, :paper
-  validates :snp_id, :uniqueness => { :scope => [:paper_id, :paper_type]}
+  validates :snp_id, uniqueness: { scope: [:paper_id, :paper_type]}
 end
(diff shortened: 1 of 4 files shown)


tsujigiri:
I feel like this should be a migration instead of a rake task. We won't do this more than once.

db/structure.sql
```diff
@@ -1696,59 +1864,52 @@ CREATE INDEX index_friendly_id_slugs_on_sluggable_type ON friendly_id_slugs USIN
...
-
-CREATE UNIQUE INDEX index_genotypes_by_snp_on_snp_name ON genotypes_by_snp USING btree (snp_name);
-
-
---
--- Name: index_snp_references_on_paper_id_and_paper_type; Type: INDEX; Schema: public; Owner: -; Tablespace: 
+-- Name: index_snps_chromosome_position; Type: INDEX; Schema: public; Owner: -; Tablespace: 
 --
 
-CREATE INDEX index_snp_references_on_paper_id_and_paper_type ON snp_references USING btree (paper_id, paper_type);
```
tsujigiri: Woops!

philippbayer: So the uniqueness constraint somehow kills the associated index. weird.

tsujigiri:
You lost the indexes on snp_references. By "unique constraint" I was referring to a unique index on all three columns. Maybe even as a primary key. The structure.sql also contains lots of stuff from your local db. You can fix this by checking out the current structure.sql from master, run `rake db:reset db:migrate` and commit the resulting structure.sql.

philippbayer:
You mean as in

     add_index :snp_references, [:snp_id, :paper_id, :paper_type], :unique => true

and remove the line in app/snp_reference.rb?

db/structure.sql
tsujigiri: The reason for this is that `SELECT ... INTO ...` doesn't care for the indexes.

tsujigiri:
By copying the table you lose the indexes. You have to re-create them. The validation on the model can stay.

philippbayer:
Aaaah now I got you, changing

philippbayer:
OK, got the master structure.sql, renamed and renewed snp_references indexes

app/models/snp_reference.rb
```diff
@@ -3,4 +3,5 @@ class SnpReference < ActiveRecord::Base
   belongs_to :snp
   belongs_to :paper, polymorphic: true
   #validates_presence_of :snp, :paper
+  validates :snp_id, :uniqueness => { :scope => [:paper_id, :paper_type]}
```
tsujigiri: The hound is supposed to prefer the colon-syntax over the hash-rocket-syntax... Stupid dog.

tsujigiri:
Apart from the syntax-nit-pick, looks good to go.

PR #208, opened 2015-10-29
Dockerize

### LINKED ISSUE #95: Configuration management
In order to not have a separate text file for everything that needs to be configured, let's put it all in a single file and make it accessible through a single mechanism.

E.g.: [configatron](https://github.com/markbates/configatron)

Additionally, let's provide a sample config, that is sufficient to make the tests run.

### LINKED ISSUE #181: replace config files with sane defaults for development
see https://github.com/gedankenstuecke/snpr/pull/180

tsujigiri:
This changes the app to make it deployable using Docker. Mainly, this means:

* Configuration is now done via environment variables (see `.env.example` and copy to `.env` for development)
* Logging in production environment goes through STDOUT 

The associated Dockerfile can be found [here](https://github.com/tsujigiri/opensnp.org-docker).

Closes #95
Closes #181


philippbayer:
This also closes #95 and #181 I assume?

(Also nice standardization of " vs. ' :neckbeard: )

tsujigiri:
Indeed it does. Updated the message.

tsujigiri:
I added the environment variable needed to push to docker hub to [travis](https://travis-ci.org/gedankenstuecke/snpr/settings).

README.md
```diff
@@ -38,13 +38,7 @@ rake db:seed OR rake db:setup (which also sets up the entire db)
...
 
-cp config/app_config.yml.example config/app_config.yml
-```
-
-## Generate Secret Token
-
-```
-bundle rake secret > secret_token
+cp .env.example .env
 ```
```
philippbayer: Does it have to be a dot-file? I feel like new collaborators may not see the file for a while (or, maybe have a line in README saying that all the setup stuff is in .env)

tsujigiri: You are looking at this line. But I can add an explanation as well. The dotenv gem expects it to be named `.env` to read and load it.

philippbayer: A short explanation in the next line would be best I think!

tsujigiri:
I added an explanation. Is this good or do you think we also need to keep the commands, so people don't just skip over it, assuming it's only text, what could possibly be informatice about it? ;)

philippbayer:
I think that's enough :)

PR #210, opened 2015-11-04
Trigger Docker Hub build of dev image

tsujigiri:
This will trigger a build of the dev image on Docker Hub. This way we don't need to build it locally and only need to run `docker run opensnp/rails-dev` locally. To run it with the version in your current working directory, run docker with `docker run -v .:/home/app/snpr`.


tsujigiri:
Nevermind, we can also trigger the dev build from the production build on Docker Hub. :+1:

PR #211, opened 2015-11-06
Fixes bug with unreplaceable avatar. Removes obsolete rake tasks. Rem…

philippbayer:
…oves solr from serverscript. Slightly refactors edit page.

Avatar was down to two things:
- wasn't accessible in the first place
~~- Paperclip and javascript based upload don't really like each other that much, just removed remote: true and that fixed it~~
- I've explicitly added the authenticity token to be sent along with the remote form, that fixes it (better than prior removal of remote form even though it didn't seem to result in any errors)

I've also changed some minor things:
- removed two obsolete rake tasks
- removed all these clearfixes that are good for nothing, obsolete overhang from old bootstrap afaik


Gemfile.lock
```diff
@@ -451,6 +451,3 @@ DEPENDENCIES
   webmock
   whenever
   will_paginate
-
-BUNDLED WITH
-   1.10.6
```
philippbayer: Where does this come from?

tsujigiri: If your bundler removes it, it means it's an old version. `gem install bundler` helps.

app/models/user.rb
```diff
@@ -2,14 +2,14 @@ class User < ActiveRecord::Base
...
-    styles: { medium: "300x300>", thumb: "100x100>#", head: "32x32#" },
+    styles: { medium: '300x300>', thumb: '100x100>', head: '32x32#'},
     default_url: 'standard_:style.png'
 
   before_validation :clear_avatar
 
   validates_attachment_size :avatar, less_than: 1.megabyte
   validates_attachment_content_type :avatar,
-    content_type: ['image/jpeg', 'image/png', 'image/gif']
+    content_type: %r/\Aimage\/.*\Z/
```
tsujigiri: Shouldn't the `\Z` be a `\z`? Also, if you use something else than `/` as a delimiter for the regex, you don't have to escape the `/` in the middle.

app/views/users/_edit.html.erb
```diff
@@ -14,37 +14,34 @@
...
-	
-	<div class="clearfix">
-		<%= f.label :name %>
-		<%= f.text_field :name %>
-	</div>
+		<%= f.file_field :avatar %>
+
 	
-    <h5><%=link_to "Change eMail/Password",{ :controller => "users", :action => "changepassword",:id => @user.id }%></h5>
+    <h5><%= link_to "Change eMail/Password", {controller: "users", action: "changepassword", id: @user.id} %></h5>
```
tsujigiri: Spaces look good after `{` and before `}`. Hound should pick that up, but apparently it doesn't like to read HTML.

tsujigiri:
Good catch with the authenticity token! I didn't even realize it was a remote form. Why would it be... 

Apart from the things I mentioned above, looks good to me! :+1:

app/models/user.rb
philippbayer: I just took it from the paperclip manual :P May as well revert it back to jpeg/png/gif, we shouldn't get other pictures anyway

philippbayer:
Yeah, why is it a remote form in the first place? Avatar-upload also works by removing the remote: true part, and it seems that the rest of the form keeps working

app/models/user.rb
tsujigiri: Ah, I see. The documentation says: "\Z - Matches end of string. If string ends with a newline, it matches just before newline". TIL.
Regarding the image type: let's be open, as in "OpenSNP"! :stuck_out_tongue:
Leaving it as is is fine. I'm just saying `%r(\Aimage/.*\Z)` is one character less for the reader to parse.

tsujigiri:
Does file upload via remote form even work? Last time I checked, it didn't. ¯\_(ツ)_/¯

philippbayer:
Well, apparently it does :D Unless we've encountered Heisencode, and now it has collapsed due to our observation.

tsujigiri:
Did I mention that it looks good to me? :+1:

philippbayer:
:+1

PR #212, opened 2015-11-08
Fix Sidekiq

tsujigiri:
In which I prepare Sidekiq for imminent reinstatement.

After deploying this and before actually running Sidekiq again, the `mailers` (and `mailnewgenotypes`?) queues should be emptied in order to prevent the emails, enqueued when testing the alternative parser, from being sent.


philippbayer:
:+1:

Issue #209, opened 2015-10-29
Foreign key constraints


tsujigiri:
In order to improve consistency in the database we should add foreign key constraints. There is even [a handy migration function for that](http://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-add_foreign_key). For that to work out, the data needs to be checked and cleaned up where necessary before, though.

philippbayer:
So far we only have a foreign key constraint between UserSnp and Snp via snp_name.

We'd need to add one for each connection - so between SnpReference and both Snp and type of the paper via snp_name, between UserAchievement and User via user_id, via Genotype and UserSnp via genotype_id, UserPicturePhenotype and User via user_id, and all the comments and their respective source via IDs.

Does add_foreign_key do any checks when you add the keys for the first time? Or do we have to write a migration that deletes any dangling connection first? I'd assume it's the second

P.S.: As a generic Rails question - so foreign keys are an additional optional method to make has_many/has_one connections more stringent so we don't have dangling DB-entries? Or why do we add them?

tsujigiri:
They are a database/Postgres thing and ensure consistency on a database level, which Rails sometimes cannot provide. Especially in environments where you have more than one instance running. And yes, we would need to clean the data before we add the constraints, otherwise postgres will complain when creating the constraints (and not do it). The question is basically, do we want to create more work for us in order to have consistent data at all times, or are we ok as it is.

philippbayer:
I'm for fixing it first - the database is going to be smaller and we can guarantee that there's no potentially private data left dangling in the database from deleted users/genotypings/crashed jobs

philippbayer:
Has been merged

Issue #143, opened 2015-03-18
Unify comments


tsujigiri:
Currently there are three models for commenting different other models. In order to decrease complexity and duplication, let's unify these to one `Comment` model with a polymorphic association to `commentable`. Let's also clean up the views.

**Write tests!**

tsujigiri:
I started working on this: https://github.com/tsujigiri/snpr/tree/one_model_to_comment_them_all

philippbayer:
Just stumbled over this nice blog-entry: http://ivory.idyll.org/blog/2015-using-hypothesis-on-workshop-sites.html

tl;dr: bind in a piece of javascript from hypothes.is, you get a new menu on the side, users can highlight sections and comment on them specifically

That would enable us to

a) have unified comments (comment *all* the things)
b) have comments on very specific parts of the site - for example, a user found out that the genotypes listed for rs1234 are flipped and should actually be A/T, or a position is wrong, or my avatar looks stupid, or ....

Also, annotations common to pieces of text are shown wherever the piece of text is shown (not sure whether we have this case)

tsujigiri:
While this is nice and generic thing, I'm not sure about the not being in control of the comments as well as having the users have to register another account in order to be able to comment. It's a pretty big step in terms of usability.

gedankenstuecke:
I agree with @tsujigiri on this one. Having to register for another account in order to be able to write comments seems too much effort for most people I'd argue. Especially as the comments are already not used too much.

gedankenstuecke:
Maybe we could tackle this problem soonish. It seems the comments are broken right now. I tried writing a SNP comment and it crashed with a 500 error in production :wink:

PR #217, opened 2015-11-19
Adds migration that removes possibly dangling DB entries

philippbayer:
In preparation for #209 this removes dangling UserSNPs, Homepages, UserPhenotypes, UserAchievements, UserPicturePhenotypes

Things that aren't dangling (I checked):
- All SNPs have UserSNPs
- All Genotypes have existing Users
- All FitbitProfiles have Users


db/migrate/20151119070640_remove_dangling_entries.rb
```diff
@@ -0,0 +1,22 @@
+class RemoveDanglingEntries < ActiveRecord::Migration
+  def up
+    UserSnp.find_each do |us|
+      if Snp.find_by_name(us.snp_name).nil? || Genotype.find_by_id(us.id).nil?
+        UserSnp.destroy(us.id)
+      end
+    end
```
tsujigiri: When importing user-SNPs, we create SNPs if they should not already exist. When there are user-SNPs without corresponding SNPs, wouldn't that be due to faulty data? Are there cases where we would delete SNPs and want the associated user-SNPs to be gone as well? Also, did you try this locally? I'm not sure this works since UserSnps don't have an `id` column.

philippbayer: It works due to some Rails magic - this is the command that is actually triggered for a UserSNP for the non-existing SNP-name "blu":

 `SQL (0.3ms)  DELETE FROM "user_snps" WHERE "user_snps"."genotype_id" = $1 AND "user_snps"."snp_name" = $2  [["genotype_id", 4], ["snp_name", "blu"]]`

But you're right, there's no ID, I assume that it can't find the ID, and then just hands the whole object to the destroy method 

I assume that we have some usersnps hanging around because sidekiq crashed/the parsing crashed, from what I've seen this used to happen, these *should* get deleted but let's just make sure

tsujigiri: Ah, I suppose this works because of the `composite_primary_keys` gem. Nice. :)

I'm all for deleting user-SNPs without genotypes, but I'm not sure we should delete the ones without SNPs, but rather create the SNPs instead.

philippbayer: Let me check whether we even have userSNPs without genotypes - if we don't have any, then we don't have to bother to delete them. It may be a bit of a hassle to find out why a specific dangling user-SNP is dangling in the first place, whether we're deleting data that a user has requested to delete and that hasn't been deleted for some reason (crash), or whether we're deleting data that a user has added to the site but that hasn't been properly linked (crash)

tsujigiri:
How about adding the foreign keys right into the same migration, to avoid having to do it twice? :)

philippbayer:
Yes, wait with the merging for now - will add soon

philippbayer:
argh forgot that this branch still had this old PR open. sheesh. but not a big change in mediawiki, tests fixed on my end

philippbayer:
Totally forgot about this PR - it has already been 2 months since the last push!

I've added a second migration that adds the foreign keys, what do you guys think?

gedankenstuecke:
I'm a bit lost on what the goal is again tbh :D

philippbayer:
Have foreign key constraints which force links between tables to always exist, impossible to have dangling links

gedankenstuecke:
In principle yes, but this is an issue with comments/messages. IIRC we had this some years back when the first users deleted their accounts and we had dangling comments/messages and we tried to render the user.name of deleted users. 

So, should the comments be deleted once a user deletes their account or should we keep them and just render "deleted user"?

philippbayer:
Hmm maybe we should even ask users when they click "Delete", some people want to have their comments deleted (personal information?), some don't (when they found some additional information etc.)

The other foreign keys are more for data consistency/sanity reasons.

gedankenstuecke:
That's a good idea, could be part of #143? For the other things I agree. Seems to me also that it just has to have a user_id but this doesn't have to be valid, right? So :+1: to merge.

PR #221, opened 2015-12-09
Added HTML emails

gedankenstuecke:
This should give us HTML email capability as well. This also brings the new email footer which includes links to Seven Bridges and Patreon for their support. Please, someone do a spell check etc. :joy:


philippbayer:
I can't see the footer - am I blind?

YixuanFranco:
hahahaha...

gedankenstuecke:
That's because I'm an idiot and forgot to add those files to the commit, I just pushed the changes in the text emails. Shouldn't do late night commits w/o :coffee: :joy:

tsujigiri:
:sunglasses:

gedankenstuecke:
Fixing the tests like a pro :sunglasses: :joy: Would be cool if someone tries sending out some test mails in their development environment. My rails installation is somehow broken (or hiredis actually) and I haven't gotten around to install the docker images ;)

gedankenstuecke:
Ok, by now also the plain text emails should have to correct footer linking to 7B and to our Patreon campaign. :smile: Having had a :coffee: makes a difference, who would have guessed!

tsujigiri:
:+1:

gedankenstuecke:
Once I get my rails back up running (or figure out how to use all that Docker setup stuff ;-)) I will give it a try. Or one of you could do it. :pray:

gedankenstuecke:
Ok, so I managed to test those emails at least partially on my system. I used http://mailcatcher.me/ to deliver mails in the development environment, which you might want to check out for your dev-systems as well. In any case: It looks like the HTML mails are now sent as expected and include the footer. 

Again: Would be great if one of you could agree on this, so we can merge this and maybe even could send out the newsletter over the weekend/on Monday? :wink:

tsujigiri:
Sound good to me! :)

gedankenstuecke:
@tsujigiri cool, could you also test the mails on your end? :smile:

tsujigiri:
I didn't get to test it, but if it worked for you, I'd assume it works, no? ;)

gedankenstuecke:
You would assume, but before sending out 5000 broken emails I'd rather double check :D

config/environments/development.rb
```diff
@@ -12,6 +12,8 @@
   config.action_controller.perform_caching = false
 
   # Don't care if the mailer can't send
+  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.smtp_settings = { :address => "localhost", :port => 1025 }
```
philippbayer: Better delete this ;)

app/views/user_mailer/welcome_user.html.erb
```diff
@@ -0,0 +1,33 @@
...
+      Cheers,
+      your openSNP-team.
+    </p>
+    --<br/>
+    openSNP thanks <a href="https://www.sbgenomics.com/">Seven Bridges</a> –
+    and its biomedical data analysis platform –
+    along with <a href="patreon.com/openSNP">our Patreons</a> for supporting our
+    hosting costs.</br>
+    <a href="https://www.sbgenomics.com">
+      <%= image_tag("http://"+ActionMailer::Base.default_url_options[:host]+"/assets/SBG_logo.png") %>
```
philippbayer: Can we just hardcode this path to opensnp.org? Even with a changed localhost path I get problems with actionmail

config/environments/development.rb
gedankenstuecke: why do you think that, it's only in the dev environment, not in production? :)

philippbayer:
After deleting the smtp localhost thing for MailCatcher, and after possibly hardcoding the HTML picture's path (I'm sure that'll break), I think it's good to go?

We may need a `rake send:newsletter:test` as a mini-rake task which just runs the same task for User ID 1 or 23

app/views/user_mailer/welcome_user.html.erb
gedankenstuecke: line 44 here should take care of it? https://github.com/gedankenstuecke/snpr/blob/master/config/application.rb

config/environments/development.rb
philippbayer: aaah yes you're right, then it would be good to stay

gedankenstuecke: yeah, i think it doesn't hurt to keep it in. :)

gedankenstuecke:
Doing a small test send sounds reasonable. But I'm somewhat opposed to hardcode the URLs inside each template, as it doesn't make any sense. For localhost make sure to add the ports :D  If you change https://github.com/gedankenstuecke/snpr/blob/master/config/application.rb line 44 to localhost:3000 it should work (at least did on my end :))

app/views/user_mailer/welcome_user.html.erb
philippbayer: Yes, it hopefully should - for me the path is `opensnp.org/assets/SBG_logo.png` which is correct, just being paranoid

philippbayer:
Yes, it works fine on localhost! Do you think a test-newsletter would be hard to pull off?

gedankenstuecke:
Nope, it's just a matter of doing UserMailer.newsletter(u).deliver_later with u being a user-object. So User.find(1) or User.find(23) should do the trick?

app/views/user_mailer/welcome_user.html.erb
gedankenstuecke: that should be the correct path once it's deployed :wink:

philippbayer:
Yes after deployment that should do the trick!

gedankenstuecke:
omg! *drum roll*

PR #232, opened 2015-12-31
Update gems

tsujigiri:
* The recaptcha gem reads the environment variables itself now, so the initializer isn't needed anymore
* Airbrake 5 introduces breaking changes, so I pinned it to '~> 4.0' for now


gedankenstuecke:
Thanks! :+1:

PR #233, opened 2015-12-31
Add test for genotype deletion

tsujigiri:


philippbayer:
Thank you for your work, Happy New Year!

PR #236, opened 2016-02-02
new newsletter for survey

gedankenstuecke:
For the follow up, as discussed via email. Would be great if someone could merge and then run the rake task :smiley_cat:


philippbayer:
Hmm so the database isn't seeded with the achievements in tests anymore?

Issue #237, opened 2016-02-05
Adding a Code of Conduct


gedankenstuecke:
I think we should add one. Even if our team is small and I think we didn't have any issues with this so far it would send a clear sign of what's (un)acceptable behaviour for us. I personally like the [Slidewinder CoC](http://www.slidewinder.io/docs/01_code_of_conduct.html). 

Any opinions on this?

tsujigiri:
Sounds like a good idea. How about having it apply to the users as well, not only to the team?

gedankenstuecke:
You mean as a CoC for opensnp.org itself instead of only using it for development? Sounds great to me! :+1:

philippbayer:
Both sound good to me!

PR #238, opened 2016-02-05
added first CoC draft

gedankenstuecke:
This could be a fix for issue #237. Directly modified from slidewinder.

### DIFF
--- a/app/views/news/index.html.erb
+++ b/app/views/news/index.html.erb
@@ -40,9 +40,13 @@
         }).render().setUser('opensnporg').start();
       </script>
     </div>
+    <h5> 2016-02-05: Adopting a Code of Conduct</h5>
+    <p>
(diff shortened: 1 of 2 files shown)


gedankenstuecke:
Uh, has someone an idea why this test fails now? :confused:

gedankenstuecke:
Ok, so just rerunning it worked in the end :recycle:

gedankenstuecke:
I talked to Kay during the Working Open Workshop and she agreed to be the external person people can talk to if any of us is the issue. :+1:

Blahah:
:+1: great issue!

philippbayer:
I think the tests still fail due to seeds sometimes not properly loading, so that achievements are missing and the relevant tests crash - it's a bit weird and I haven't yet figured out why that happens

The text itself looks good to me; we may need one more "external" person at some point, but I think that our community isn't large enough to warrant that right now.

+1 for merge

gedankenstuecke:
Just rerunning it worked though, so that might not be the issue with the tests. 

Regarding a second external person: I think that's fine for now. Once we're larger we can get another person of authority. :smile: :+1:

PR #251, opened 2016-02-10
Fixes undefined method [] for nil:NilClass error in ALM details

philippbayer:
It seems that the output of the ALM API changed a little bit, causing heaps of `NoMethodError: undefined method [] for nil:NilClass`

from this line:

`readers_total = detail_result[0]["views"]`

because it's not a list anymore, so [0] gives you nil, so you can't run ["views"] on that

This commit puts in an example reply in the comments (try it for yourself: http://alm.plos.org/api/v3/articles/10.1371/journal.pone.0080003?api_key=OUR_API_KEY ) and fixes this bug, and adds a little check to see whether the HTTP response from the API is correct so that we don't parse garbage or empty text.

~~Edit: This commit also fixes that annoying failing UserAchievement test~~

Edit 2: Apparently the above link works without the API-key, too. huh


philippbayer:
Hm the UserAchivement test still fails:      

    Failure/Error: expect(UserAchievement.find_by(id: user_achievement.id)).to be_nil

gedankenstuecke:
For some reason this only fails sometimes. Have you tried just rerunning the build on Travis? I think this might be some bug in the testing instead?

--
Sent from a mobile device

While I may be sending this email outside my normal office hours, I have no expectation to receive a reply outside yours

philippbayer:
It looks like the achievements are not properly loaded. 

The first commit bb4372c fixes this by force-creating the achievement that test needs, but it's not a real solution for other tests. In the second commit I removed that and force RSpec to load the achievements, which works perfectly fine on my system. Rebuilding now

app/workers/plos_details.rb
```diff
@@ -20,15 +20,33 @@ def perform(plos_paper)
                  "api_key=#{ENV.fetch('PLOS_API_KEY')}"
 
     detail_resp = Net::HTTP.get_response(URI.parse(detail_url))
+    # did we get a proper response?
+    unless detail_resp.is_a? Net::HTTPSuccess
+      # don't bother parsing broken or empty output
+      sleep 6
```
tsujigiri: Instead of the `sleep` calls, should we give [sidekiq-throttler](https://github.com/gevans/sidekiq-throttler) a try? Seems cleaner to me and would keep the job from unnecessarily blocking the worker.

tsujigiri:
A test would be great! :)

app/workers/plos_details.rb
philippbayer: That looks great and would be useful for all API-jobs! How about we put that into a second PR?

philippbayer:
Isn't that handled in test/unit/plos_search_test.rb?

app/workers/plos_details.rb
tsujigiri: Or even better, [configuring the retry delay](https://github.com/mperham/sidekiq/wiki/Error-Handling#configuration):

```ruby
class WorkerWithCustomRetry
  include Sidekiq::Worker
  sidekiq_options :retry => 5

  # The current retry count is yielded. The return value of the block must be 
  # an integer. It is used as the delay, in seconds. 
  sidekiq_retry_in do |count|
    10 * (count + 1) # (i.e. 10, 20, 30, 40)
  end

  def perform(...)
  end
end
```

tsujigiri:
This one only tests `PlosSearch`. If you want, I can do it.

app/workers/plos_details.rb
philippbayer: Doesn't that fire only with the first retry?

philippbayer:
I can do it too :) High time to write some tests for this one

gedankenstuecke:
:+1:

tsujigiri:
Great! :)

app/workers/plos_details.rb
tsujigiri: Yes. We would need to make the worker throw an exception when the response is about the request limit being reached. It would then retry after a certain time. Do we know what exactly the rate lmit is? 10 per minute? Then let's just make the delay 30 seconds.

gedankenstuecke: From the PLOS API: 

> PLOS Search API rate limit: Please limit your API requests to 7200 requests a day, 300 per hour, 10 per minute and allow 5 seconds for your search to return results. If you exceed this threshold, we will lock out your IP address. If you’re a high-volume user of the PLOS Search API and need more API requests a day, please contact us at api@plos.org to discuss your options. We currently limit API users to no more than five concurrent connections from a single IP address.

tsujigiri: Hm... OK. If they actually lock you out once you exceed the limit, `sidekiq-throttler` might be the better option. ;)

spec/features/delete_genotype_spec.rb
```diff
@@ -2,8 +2,8 @@
   let(:user) { create(:user) }
   let!(:genotype) { create(:genotype, user: user) }
   let!(:user_achievement) do
-    create(:user_achievement, achievement: Achievement.find_by(award: 'Published genotyping'),
-                              user: user)
+    create(:user_achievement, achievement_id: Achievement.find_by(award: 'Published genotyping').id,
+                              user_id: user.id)
```
tsujigiri: Doesn't the test rather fail due to the missing seeds? Whether you use the instance or the id shouldn't make any difference, except that it looks nicer using the instances. ;)

spec/spec_helper.rb
```diff
@@ -13,6 +13,8 @@
 
 Sidekiq::Logging.logger = Logger.new('log/sidekiq-test.log')
 
+load "#{Rails.root}/db/seeds.rb"
```
tsujigiri: This shouldn't be necessary. Running `bundle exec rake db:reset` should do the trick.

philippbayer: But why does it fail on Travis-CI with the same error?

tsujigiri: Very good question.

tsujigiri: I created a separate issue for this. #253

philippbayer: You want me to take this force-fix out for this PR?

tsujigiri: If the tests break otherwise, leave it. We can remove it when we found out why we need it.

philippbayer:
This now fails with "Real HTTP connections are disabled", of course. How do I do this best in RSpec? Should I implement the whole VCR thing that's set up for test/unit?

tsujigiri:
VCR is for integration tests (see `spec/integration`). In a unit test you should just stub out everything that doesn't belong to the class under test. Something along the lines of:

```ruby
response = double('response', body: '{"some":"JSON"}')
expect(Net::HTTP).to receive(:get_response).with('http://blah.blah').and_return(response)
```

philippbayer:
Now it works!! Thanks @tsujigiri

spec/integration/plos_details_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper.rb'
+
+describe PlosDetails, :vcr do
+  let(:plos_paper) do
+    create(:plos_paper, id: 1, reader: 0, doi: '10.1371/journal.pone.0089204')
+  end
+  let(:job) {PlosDetails.new}
```
tsujigiri: You don't need this line if you use `subject` instead of `job` in the test. As a side-note: You can override the default `subject` by calling it with a block on the same level as the `let`:

```ruby
subject { described_class.new(:foo, :bar) }
```

spec/integration/plos_details_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+
+  it 'updates the view count' do
+    before = plos_paper.reader
+
+    VCR.use_cassette 'plos_detail' do
+      job.perform plos_paper.id
+    end
```
tsujigiri: You don't need the `use_cassette` block, that's what the `:vcr` on top is doing, deriving the name of the VCR file from the test.

spec/integration/plos_details_spec.rb
```diff
@@ -0,0 +1,18 @@
...
+    expect(plos_paper.reload.reader).not_to eq(before)
```
tsujigiri: Since you set `reader` to 0 when setting up the paper, you don't need to check against the before here. Checking that `reader` has a value after running the job is enough. Assinging `before` is also not necessary, then.

spec/integration/plos_details_spec.rb
philippbayer: If I remove these lines I get:

           There is currently no cassette in use. There are a few ways [...]

doesn't matter whether the cassette is named plos_detail.yml, plos_details_spec.yml etc. Missing some config somewhere?

tsujigiri: See edit above. :)

tsujigiri:
:ship:

PR #254, opened 2016-02-11
added fix for parsing

gedankenstuecke:
We had a slight discrepancy between our validity checks for file parsing and our actual file parsing. While the actual parsing was done after splitting columns using `row.strip.split("\t")` the validity checks only used `row.split("\t")`.

I unified it in order to make the checks pass/fail in the right circumstances.


tsujigiri:
Looks like we should refactor this whole thing so we don't have the same logic in two places. But for now, this should do.

PR #257, opened 2016-02-13
Ignores SNPedia warnings until we update gems

philippbayer:
This just adds the ignorewarnings => true flag so that the logs aren't so huge until we implement the mediawiki-ruby-api gem (which is a hassle, no list() command as far as I can see)

(The list must be something like https://snpedia.com/api.php?action=query&list=search&srsearch="Rs1234(" )


tsujigiri:
:ship:

gedankenstuecke:
:+1:

PR #263, opened 2016-03-01
Rplots

### LINKED ISSUE #102: Add stats page
As Samantha proposed on the mailing list, we should have a stats page showing # users, # genotypings (with percentage of each provider etc.), some graphs over time etc.

gedankenstuecke:
**tl;dr**
- [x] Adds necessary code for statistics
- [x] Restructures the navbar

This merge would introduce a slightly restructured Navbar for openSNP. The Users-Item would be gone and the new "Stats"-item would appear. I also regrouped the items slightly to make a bit more sense: ((News,Stats), (Snps,Phenotypes,Genotypes)). The stats page is also added and displays the graphs generated by openSNP/growthplots. Should be good to go like this. Closes #102

### DIFF
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -20,6 +20,8 @@
           <a class="brand" href="/">openSNP</a>
           <ul class="nav">
             <li class=""><%= link_to("News", '/news') %></li>
+            <li class=""><%= link_to("Stats", '/statistics') %></li>
+            <li class="divider-vertical"></li>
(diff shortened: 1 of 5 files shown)


philippbayer:
You can also delete views/users/index.html.erb and the route to it, or do we still link there from anyhere?

gedankenstuecke:
Oh, not sure actually. I think we could link to it from the news-page?

gedankenstuecke:
I implemented the links from the news page for now. Any other things to consider before merging?

PR #277, opened 2016-04-10
Replaces foreign keys by cascade-delete ones. 

philippbayer:
- also adds user_snps queue to sidekiq worker in serverscript.

This one fixes the 500 issue currently caused on deletion of users or genotypes - by default foreign keys have :restrict on update and delete actions, so you can't just do that. This commit is an idea: it deletes the foreign keys (after all that work... apparently you can't alter PSQL foreign keys) and adds new ones on non-SNP tables which cascade on delete instead.
I've excluded the user-SNP table since the migration took a surprisingly long time.

### DIFF
--- a/db/migrate/20160410105934_remove_foreign_key_constraints.rb
+++ b/db/migrate/20160410105934_remove_foreign_key_constraints.rb
@@ -1,0 +1,23 @@
+class RemoveForeignKeyConstraints < ActiveRecord::Migration
+  def change
+    remove_foreign_key :user_snps, :genotypes
+    remove_foreign_key :genotypes, :users
+    remove_foreign_key :fitbit_profiles, :users
+    remove_foreign_key :homepages, :users
+    remove_foreign_key :user_phenotypes, :users
+    remove_foreign_key :user_picture_phenotypes, :users
+    remove_foreign_key :phenotype_comments, :users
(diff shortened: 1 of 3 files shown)


tsujigiri:
That's cool! I wasn't even aware of that feature. Rails can do that as well, though ([see `:dependent` option](http://api.rubyonrails.org/classes/ActiveRecord/Associations/ClassMethods.html#method-i-has_many)) and I'm wondering if it's better to keep this on the application level. With this change, the database knows something the application doesn't. Associated objects are being deleted. In case of the comments, I'm not sure if they even should be just deleted. Foreign key constraints are great for keeping the data consistent. If a foreign key error pops up, we know, we missed something on the application level. Having the database take care of this automatically works, but adds a place where you have to look (and know about in the first place) to fully grasp what's going on.

tsujigiri:
Also, I just checked for the foreign keys in the db, the migration never finished... :D I manually added the foreign keys now.

philippbayer:
There are several migration-related errors in the error db, so I assume that it started a new migration with each deploy and at some point they all fought and crashed ("mutually assured destruction")

tsujigiri:
I added the timestamp of the migration to the `schema_migrations` table, so it thinks it already did that. ;)

philippbayer:
#278 is way more :fireworks:, closing this one

PR #278, opened 2016-04-10
Add `dependent: :destroy`s to reflect foreign key constraints

tsujigiri:
Counter-proposal for #277

### DIFF
--- a/app/controllers/genotypes_controller.rb
+++ b/app/controllers/genotypes_controller.rb
@@ -18,6 +18,7 @@ class GenotypesController < ApplicationController
     @genotype = Genotype.create(genotype_params)
     @genotype.user = current_user
     if @genotype.valid? && @genotype.save
+      Preparsing.perform_async(@genotype.id)
       # award for genotyping-upload
       @award = Achievement.find_by_award("Published genotyping")
       user_achievement_attrs = { achievement_id: @award.id,
@@ -50,21 +51,19 @@ class GenotypesController < ApplicationController
   end
 
   def destroy
-    @user = current_user
-    @genotype = Genotype.find_by_id(params[:id])
-    if @genotype.destroy
-      flash[:notice] = "Genotyping was successfully deleted."
-      if @user.genotypes.count == 0
-        # update user-attributes
-        @user.update_attributes(has_sequence: false, sequence_link: nil)
+    user = current_user
+    genotype_count = user.genotypes.count
+    DeleteGenotype.perform_async(genotype_id: params[:id])
+    flash[:notice] = 'Your Genotyping will be deleted. This may take a few minutes.'
+    if genotype_count == 1
+      # update user-attributes
+      user.update_attributes(has_sequence: false, sequence_link: nil)
 
-        # delete Uploaded Genotyping-achievement
-        achievement = Achievement.where(award: 'Published genotyping')
-        UserAchievement.where(achievement: achievement, user: @user)
-                       .destroy_all
-      end
-      redirect_to current_user
+      # delete Uploaded Genotyping-achievement
+      achievement = Achievement.where(award: 'Published genotyping')
+      UserAchievement.where(achievement: achievement, user: user).destroy_all
     end
+    redirect_to current_user
   end
 
   private 
--- a/app/models/genotype.rb
+++ b/app/models/genotype.rb
@@ -13,23 +13,12 @@ class Genotype < ActiveRecord::Base
     size: { in: 0..100.megabytes }
   do_not_validate_attachment_file_type :genotype
 
-  after_create :parse_genotype
-  before_destroy :delete_genotype
-
   def is_image?
     false
   end
 
   def fs_filename
     "#{user.id}.#{filetype}.#{id}"
-  end
-
-  def parse_genotype
-    Preparsing.perform_async(id)
-  end
-
-  def delete_genotype
-    DeleteGenotype.perform_async(genotype_id: id)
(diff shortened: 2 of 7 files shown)


tsujigiri:
I'll fix the tests soon... I promise! :D

philippbayer:
This one looks way clearer than mine - so it would keep the foreign keys but "overwrite" the on_delete behavior in the models?

P.S.: I've never seen "Stack Level too deep" in a test :D

philippbayer:
I can confirm that with this branch, triggering genotype deletion deletes the data and no 500.

However, it looks like the click already triggers all of the deletion action - even with a small genotyping of 25kb the browser waits for a few seconds, and the DeleteGenotype job doesn't have a genotype to work with:

```2016-04-10T14:28:08.591Z 5074 TID-18ryz0 DeleteGenotype JID-fdfd0c0a44f6b2deb4a1b129 INFO: start
2016-04-10T14:28:08.595Z 5074 TID-18ryz0 DeleteGenotype JID-fdfd0c0a44f6b2deb4a1b129 DEBUG: Failure! Retry 1 in 34 seconds
2016-04-10T14:28:08.597Z 5074 TID-18ryz0 DeleteGenotype JID-fdfd0c0a44f6b2deb4a1b129 INFO: fail: 0.006 sec
[...]
2016-04-10T14:28:08.598Z 5074 TID-18ryz0 WARN: ActiveRecord::RecordNotFound: Couldn't find Genotype
2016-04-10T14:28:08.599Z 5074 TID-18ryz0 WARN: /home/drsnuggles/.rvm/gems/ruby-2.0.0-p247@opensnp/gems/activerecord-4.2.5.1/lib/active_record/core.rb:196:in `find_by!'
/home/drsnuggles/opensnp/tsujigi/snpr/app/workers/delete_genotype.rb:6:in `perform'
``` 

So it looks like we should kajiggle this to not get destroyed on dependent, but do nothing?

gedankenstuecke:
Just so I get the problem: The issue is that clicking "delete" already deletes the genotype and the worker will try to perform the same action afterwards?

philippbayer:
Yes, it looks that way!

gedankenstuecke:
Yeah, this is why we implemented the worker in the first place. So one way to get around it would be to not really delete a genotype when users press the button on the front-end but 

1. enqueue the delete job
2. mark the Genotype as "to-be-deleted"? That way the genotype object would still exist but we could check for the flag and only offer those for download etc. that are unmarked?

tsujigiri:
The idea is to run the worker instead of calling destroy on the model in the controller. I didn't get to finish that yesterday.

gedankenstuecke:
:+1: But should we add a flag for filtering what ends up in the dump etc? Otherwise it could lead to confusion if people press "delete" but the data is still displayed because the worker hasn't finished yet?

tsujigiri:
I would just change the message, so that it says "will be deleted" or maybe even more detailed. That should be good enough, right?

gedankenstuecke:
That would work too, but adding the flag would be "nicer" in a way that people who urgently want to delete their data at least will be assured that the raw data files aren't on display/in downloads any longer?

philippbayer:
Then it would be best to not have it in a popup, but to redirect to a new site with a detailed description of what's happening to which dataset (i.e.: file is gone; your user SNPs are being deleted; your user account is gone; etc)

gedankenstuecke:
> You pressed *delete genotype*. We have now marked your genotype file for deletion. This means that the file and the variants associated with it won't be displayed on the website any longer and that each genetic variant listed in your data file will also be removed from our database (this can take some time as it can be millions of entries).  
>
> The next complete dump of openSNP will also no longer contain your file (this can take up to 24 hours in the worst case) 

Something along those lines?

philippbayer:
Yes! And something very similar for "Delete user account", AFAIK that one redirects to the home-page now with a small alert.

But that is something for another PR...

tsujigiri:
If you think that's best, sure. If any of you has the time, feel free to continue what I started. otherwise I might get to it tomorrow morning.

gedankenstuecke:
@tsujigiri I think from the expected behavior for the user that would be nicest? But there might be good reasons to do it differently, In that case I'm happily convinced otherwise :-)

tsujigiri:
As always, only that it's more code to add, test and maintain. ;)

gedankenstuecke:
That's a good reason, but do you think the benefits on the user-side outweigh the drawbacks on the code-side? :)

tsujigiri:
I think if we tell the user what's going on, it should be fine. If you are concerned because the data would still show up in the dump for some time, we can also do the flag. But then we have to touch the dumping as well.

gedankenstuecke:
I wouldn't touch the dump directly, because that one is done daily in any case, but depending on the queues etc. there might be a loooong delay until a genotype is actually removed completely?

tsujigiri:
By now, our queues are mostly empty. So, it shouldn't take longer than a few minutes anymore.

gedankenstuecke:
Ok, then let's do the low maintenance version of just displaying the text :)

tsujigiri:
I fixed the tests and moved the triggering of the parsing and deleting jobs out of the hooks in the model, into the controller. This makes it less implicit and easier testable.

philippbayer:
I  can confirm that now there's no wait after user clicks "Delete genotype", and the job properly deletes it.

:+1: for merging

gedankenstuecke:
:+1:

PR #283, opened 2016-05-15
Update structure.sql

tsujigiri:
I ran `rake db:structure:dump` on production, so we are up-to-date with
the production database after adding the foreign keys and removing old
tables.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -496,18 +496,7 @@ CREATE TABLE genotypes (
     genotype_file_name character varying(255),
     genotype_content_type character varying(255),
     genotype_file_size integer,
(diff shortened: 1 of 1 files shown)


tsujigiri:
It removes migration `20151119070640` (a.k.a. `RemoveDanglingEntries`) from the `schema_migrations`, which means it didn't run on production, yet. Is that something we need to look into?

philippbayer:
IIRC, that migration was just there to make sure that the foreign keys will work out, but it looks like it didn't even delete anything.

tsujigiri:
I inserted the timestamp into `schema_migrations`, then.

Issue #285, opened 2016-05-27
Removing own user-phenotypes via "Edit your phenotypes" does nothing


philippbayer:
After a refresh of the page the phenotype is back. Found by @MateusJabour 
Firebug doesn't report any errors related to this, logs are empty too. I assume it has something to do with the new foreign keys, I'll check more this weekend.

PR #286, opened 2016-05-27
Fixes problem with user phenotype updates

philippbayer:
Turns out, the strong parameters from Rails 4 were the culprit for #285, we just never realized. Users weren't able to update user-phenotypes or remove them. This should fix this.

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -273,6 +273,7 @@ class UsersController < ApplicationController
       :message_on_new_phenotype,
       :message_on_phenotype_comment_reply,
       :message_on_snp_comment_reply,
+      user_phenotypes_attributes: [:id, :variation, :_destroy]
     )
   end
 end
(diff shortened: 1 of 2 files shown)


app/controllers/users_controller.rb
```diff
@@ -273,6 +273,7 @@ def user_params
       :message_on_new_phenotype,
       :message_on_phenotype_comment_reply,
       :message_on_snp_comment_reply,
+      user_phenotypes_attributes: [:id, :variation, :_destroy]
```
tsujigiri: I don't think the `_destroy` needs to be in there, as it's just for rails to figure out the HTTP method. Usually browsers don't support all of them, so Rails uses the additional parameter to work around that constraint.

philippbayer: That's possible, let me try it out!

philippbayer: OK I tried it now - when I delete the user-phenotype via the edit phenotype form I get this error in the logs:

`Unpermitted parameter: _destroy`

and it doesn't get deleted and reappears on refresh, so I think I'll keep it!

PR #291, opened 2016-06-02
Cleanup cruft 

philippbayer:
This PR details my work on openSNP during the Sprint. I'll clean up the place and list each commit here.

- remove stone-old js piece from Google that supposedly helps with IE8,9. We never tested it, the js piece itself is gone since who-knows-when.
- adds a mini meta tag for IE8/9, the recommended way from Bootstrap now
- the current News page has a Twitter widget, but that piece of JS still used the old obsolete Twitter API and silently disappeared. This replaces it with a new widget, but it's not perfectly pretty yet.
- ever since the move to Docker the New Relic integration hasn't worked, this removes the integration since we don't use it much anyway.
- removes some active admin overhang
- update bootstrap classes login and edit forms - Lots of ugly clearfix divs everywhere. Still some left - they're still supported by bootstrap CSS, but they're not in the docs anymore as far as I can see
- removes beautifier.rb since no-one uses it
- fixes useless "missing template: remove_help_one/two/three" error messages
- the help_one to help_three help is a bit more logically structured with less logic in the view

I'm hoping that the final PR by Friday or Saturday will end up deleting more lines than adding.

~~I'll be editing this message while I'm adding commits, so please don't merge yet :)~~
The Sprint has ended, this can be investigated and criticized


philippbayer:
Hmm OK :) I'll put them back, or even make proper resource paths, I'll have a read about it

app/controllers/users_controller.rb
```diff
@@ -222,14 +222,24 @@ def destroy
 
   def remove_help_one
     current_user.update_attribute("help_one",true)
+
+    respond_to do |format|
+      format.js {render nothing: true}
+    end
```
tsujigiri: You don't need the `respond_to` here. Just `render nothing: true` is enough. We don't call it any other way anyway.

PR #297, opened 2016-06-03
Add a missing https to dalliance

philippbayer:
My Firefox complains about `Blocked loading mixed active content "http://www.biodalliance.org/datasets/hg38.2bit"` when trying to load any snps/show page, this adds the https so that it'll work (hopefully? it's not testable on localhost)


gedankenstuecke:
:+1:

PR #300, opened 2016-06-09
fixed references to emailing us

gedankenstuecke:
At some points in the emails we send to users we still referenced that they should just reply to these emails. But we changed this at some point, so that emails are no longer from `info@opensnp.org` but from `donotreply@opensnp.org`. I replaced those parts so that users now again know how to reach us.

### DIFF
--- a/app/views/user_mailer/duplicate_file.html.erb
+++ b/app/views/user_mailer/duplicate_file.html.erb
@@ -14,7 +14,7 @@
     </p>
     <p>
       If there is no genotyping listed in your user-account, please contact us!</br>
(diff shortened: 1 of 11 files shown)


tsujigiri:
That's a lot of repitition, how about putting it into a layout, all emails would be using?

gedankenstuecke:
I guess we'd need more than one layout, but in principle: Sure, we should do that. Would still like to merge this now in order not to send out wrong emails until we did the layouts :)

tsujigiri:
Sure. Let's create another issue, then, for the layouts.

PR #302, opened 2016-06-11
Fix genotype deletion

tsujigiri:
This changes the relationship between `Genotype` and `UserSnp`, so the `UserSnp`s get deleted in the database instead of being loaded and destroyed one by one. I also moved related logic into the worker.


philippbayer:
:+1: :sparkles: :sparkling_heart: :fireworks:

Issue #304, opened 2016-06-15
add unique key constraint on Phenotype table


vivekiitkgp:
Following from discussion on gitter.

vivekiitkgp:
@gedankenstuecke I am going to assume that since we have a validation step at the controller stage there will be no duplicate entries in the `Phenotype` table right now. If not, then one may have to operate on the table and merge duplicate entries...

gedankenstuecke:
@vivekiitkgp I think there were like 2-3 duplicate phenotypes which would have to be manually dealt with before. Thanks for the work on it so far!

tsujigiri:
I'd prefer that we deal with it through a reviewable and testable migration over fiddeling around with it manually. ;)

gedankenstuecke:
Fair enough :p

vivekiitkgp:
@tsujigiri On it!

PR #305, opened 2016-06-15
Add index, unique key constraint in Phenotype table

raivivek:
Signed-off-by: Vivek Rai vivekrai.iitkgp@gmail.com
#304


philippbayer:
You may have to run `rake db:migrate` and then commit the new schema to get the test to pass

PR #306, opened 2016-06-16
easy user_snp cache fix

### LINKED ISSUE #289: "Other users" count of SNPs is broken
See for example https://opensnp.org/snps/i6000211#users where it says -10

In the database:
>Snp.find_by_name("i6000211").users.length
=> 616

So it's not broken there.

gedankenstuecke:
should fix #289 when running the rake task


philippbayer:
Looks good to me

gedankenstuecke:
Great. Shall we run the update-task once it's deployed?

PR #309, opened 2016-06-19
Schedule updating papers once a month

tsujigiri:
Instead of updating the papers for each SNP whenever their page gets requested, update the papers of all SNPs if they haven't been for a month.

### DIFF
--- a/app/workers/update_papers.rb
+++ b/app/workers/update_papers.rb
@@ -6,7 +6,7 @@ class UpdatePapers
     MendeleySearch => 'mendeley_updated',
     PlosSearch => 'plos_updated',
     Snpedia => 'snpedia_updated'
-  }
+  }.freeze
 
   def perform
     PAPER_UPDATED_COLUMNS.each do |worker, column|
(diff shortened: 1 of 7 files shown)


app/workers/update_papers.rb
```diff
@@ -0,0 +1,24 @@
+class UpdatePapers
+  include Sidekiq::Worker
+  MAX_AGE = 31.days
+
+  PAPER_UPDATED_COLUMNS = {
```
tsujigiri: :white_check_mark:

config/schedule.rb
```diff
@@ -22,3 +22,7 @@
 every :day, at: '1am' do
   rake 'dump:full', environment: 'production'
 end
+
+every :day do
```
philippbayer: shouldn't this be `:month`?

tsujigiri: I was thinking, updating all of the papers at the same time every month might be a bit heavy on the load. By running this once a day, we update all the papers that haven't been for 31 days, more or less evenly throughout the month.

gedankenstuecke: 👍  sounds good!

philippbayer: Sounds fine to me too!

PR #316, opened 2016-06-29
Limit frequency queue to 1

tsujigiri:
Since the weights don't seem to work properly and the frequency jobs still block the parsing jobs, let's reduce the number of simultaneous frequency jobs to 1.

### DIFF
--- a/config/sidekiq.yml
+++ b/config/sidekiq.yml
@@ -30,7 +30,7 @@
   zipgenotyping: 1
   zipfulldata: 1
   fitbit: 3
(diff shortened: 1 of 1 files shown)


tsujigiri:
If the weighting doesn't work prperly anyway, maybe we should try the priority approach instead. If the queues have no weight, they are polled in the order they appear in the list. At least that's what I get from the [docs](https://github.com/mperham/sidekiq/wiki/Advanced-Options). But for now, let's merge this.

PR #318, opened 2016-07-17
decreases calls to DB when vieweing phenotypes index

audy:
Indexing phenotypes is slow. I believe this is due to looping over the entire table at least once to build a JSON object which I don't think is even being used by the html or xml views.

- Moved the JSON generating code within the json block so it doesn't run when rendering HTML
- Changed the `each` loop to a `find_each` loop used in generating the JSON. This causes phenotypes to be fetched in batches.
- Eager load `user_phenotypes` to reduce DB calls.

### DIFF
--- a/app/controllers/phenotypes_controller.rb
+++ b/app/controllers/phenotypes_controller.rb
@@ -3,11 +3,11 @@ class PhenotypesController < ApplicationController
   helper_method :sort_column, :sort_direction
 
   def index
-    @title = "Listing all phenotypes"
+    @title = 'Listing all phenotypes'
 
-    @phenotypes = Phenotype.
-      order(sort_column + " " + sort_direction).
-      includes(:user_phenotypes)
+    @phenotypes = Phenotype
+      .order(sort_column + ' ' + sort_direction)
+      .includes(:user_phenotypes)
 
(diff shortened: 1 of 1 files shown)


app/controllers/phenotypes_controller.rb
```diff
@@ -3,22 +3,29 @@ class PhenotypesController < ApplicationController
...
       format.xml 
-      format.json {render :json => @phenotypes_json} 
+      format.json do
+        phenotypes_json =
+          @phenotypes.find_each.map do |p|
+            phenotype = {}
+            phenotype['id'] = p.id
+            phenotype['characteristic'] = p.characteristic
+            phenotype['known_variations'] = p.known_phenotypes
+            phenotype['number_of_users'] = p.user_phenotypes.length
```
tsujigiri: The building of the hash could also be improved:

```ruby
{
  id: p.id,
  characteristic: p.characteristic,
  known_variations: p.known_phenotypes,
  number_of_users: p.user_phenotypes.length
}
```

audy: I was just thinking that this would get fixed when the switch to jbuilder is made.

tsujigiri: Fair enough.

PR #319, opened 2016-07-17
seeds development db with some phenotypes

audy:
Running `rake db:seed` will not create 100 phenotypes. Useful for testing.

### DIFF
--- a/db/development_seeds.rb
+++ b/db/development_seeds.rb
@@ -1,0 +1,37 @@
+BODY_PARTS = %w{
+  hair
+  head
+  eyebrow
+  eyelash
+  eye
+  ear
+  nose
+  nostril
+  mouth
+  teeth
+  tongue
+  forearm
+  hand
+  finger
+  fingernail
+  thumb
+  torso
+  buttocks
+  leg
+  knee
+  heal
+  foot
+  toe
+  toenail
+}.freeze
+
+PROPERTIES = %w{
+  length
+  width
+  color
+  size
+}.freeze
+
+100.times do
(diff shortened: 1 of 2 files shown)


tsujigiri:
Great idea. I'm just not sure whether random values belong in the factories. I guess the tests should be deterministic, unless you are explicitly doing some kind of property testing. Funnily the one constant is named `PROPERTIES`, so one could argue that it indeed is some kind of "property testing"... :grimacing: The `PROPERTIES` and `BODY_PARTS` constants could be moved to the `seed.rb`, but the seeds are used for the tests as well, so that wouldn't really change anything. Since these are meant for development and not tests, maybe we should create a separate seed file for development? What do you think?

audy:
> I guess the tests should be deterministic, unless you are explicitly doing some kind of property testing

Now that you mention it, this can cause some problems in testing if you have a uniqueness constraint and the random value collides when an existing record.

> Since these are meant for development and not tests, maybe we should create a separate seed file for development?

It's up to you. I can see why you wouldn't want to mix seed and test code. Let me know and I'll move it.

tsujigiri:
I think it should work if you load a separate seed file within the existing one, but only in the dev environment. Something like this:

```ruby
load 'development_seeds.rb' if Rails.env.development?
```

tsujigiri:
Can you also move `PROPERTIES` and `BODY_PARTS` into `development_seeds.rb`? Than I'm happy. :)

audy:
That was unintentional.

tsujigiri:
:thumbsup:

PR #320, opened 2016-07-17
remove Sidekiq migration

audy:
You don't need this anymore and it causes a minor hiccup when setting up the app for development (you need to be running Sidekiq which is unnecessary when you're developing)


tsujigiri:
You don't need to run the migrations in order to set up the app. `rake db:setup` is enough, which runs `db:schema:load`, which loads the schema from the `db/schema.rb`. In general, migrations are not supposed to work forever, since the code they rely on might change.

audy:
> `rake db:setup` is enough, which runs `db:schema:load`

Wow I did not know that. Never mind I guess.

PR #322, opened 2016-07-17
removes username and password from DB

audy:
These are usually blank by default. Adds an extra step when getting setup for development.


tsujigiri:
:thumbsup:

PR #323, opened 2016-07-18
adds specs for phenotypes json API

audy:
I wrote some specs for the phenotypes API based on the wiki documentation. They're incomplete but better than no tests 😉


tsujigiri:
:clap: :clap: :clap: :clap: :clap:

PR #324, opened 2016-07-18
reduces calls to DB for indexing genotypes

audy:
- pre-fetch users
- remove call to `Genotype.all` in index view which was causing `n_row` calls to `Genotype.all`. Do you really need the `#` column? If so, I can use `#each_with_index` or something instead.

### DIFF
--- a/app/controllers/genotypes_controller.rb
+++ b/app/controllers/genotypes_controller.rb
@@ -5,8 +5,11 @@ class GenotypesController < ApplicationController
 
   def index
     @title = "Listing all genotypings"
-    @genotypes = Genotype.order("#{sort_column} #{sort_direction}")
-    @genotypes_paginate = @genotypes.paginate(page: params[:page],per_page: 20)
+    @genotypes =
+      Genotype
+        .includes(:user)
+        .order("#{sort_column} #{sort_direction}")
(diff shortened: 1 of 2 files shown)


app/controllers/genotypes_controller.rb
```diff
@@ -5,8 +5,11 @@ class GenotypesController < ApplicationController
 
   def index
     @title = "Listing all genotypings"
-    @genotypes = Genotype.order("#{sort_column} #{sort_direction}")
-    @genotypes_paginate = @genotypes.paginate(page: params[:page],per_page: 20)
+    @genotypes =
+      Genotype
+        .includes(:user)
```
audy: I disagree.

app/controllers/genotypes_controller.rb
```diff
@@ -5,8 +5,11 @@ class GenotypesController < ApplicationController
...
+        .order("#{sort_column} #{sort_direction}")
```
audy: Which planet are you from?

tsujigiri:
:+1:

app/controllers/genotypes_controller.rb
tsujigiri: Yeah, me, too. We should adjust that in the config.

PR #325, opened 2016-07-19
cleans up user_is_user template

audy:
I was going to try cleaning up the users controller but ended up going on a hunt for a missing `<div>` instead. I ended up cleaning up the `_user_is_user` partial quite a bit. I mostly just fixed indentation and replaced some inline Ruby code with more concise equivalents.

I tested in development mode by clicking on stuff and everything seems to be in order.

### DIFF
--- a/app/views/users/_user_is_user.html.erb
+++ b/app/views/users/_user_is_user.html.erb
@@ -1,198 +1,231 @@
 <div class="row">
-    <div class="span3"><h1>Hello <%= @first_name %>!</h1></div>
-    <div class="row">
-        <br/>
-        <div align="right">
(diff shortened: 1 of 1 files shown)


tsujigiri:
:honeybee:

tsujigiri:
A test would still be nice, though. If you have the time. :wink:

audy:
What kind of test? A view test? I usually just test that the page renders without raising an error when I don't have a lot of time.

--austin

tsujigiri:
To increase test coverage I do the same. A simple feature test that opens the page should be enough. There were some issues. You can see them if you look into Gitter. ;)

PR #330, opened 2016-08-12
Snp update feed

philippbayer:
This implements a rudimentary 'What's happening to your SNPs' feed for logged-in users who have uploaded SNPs.

Not finished yet - ~~test doesn't work.~~ ~~getting SNPs is extremely slow since Rails runs an INNER JOIN on SNPs + userSNPs, that takes about 20-30s. Thinking of better ways to do it~~ I've now made it in a slightly weird way - instead of getting all SNPs for the user, I'm getting the last 30 updated papers for each type and check whether their SNPs's list of users  contains the current_user. It's quite a bit faster at the expense of accuracy, there's a small chance that a user has no updated papers


app/controllers/users_controller.rb
```diff
@@ -112,6 +111,9 @@ def show
     @phenotype_comment_replies.sort! { |b,a| a.created_at <=> b.created_at }
     @paginated_phenotype_replies = @phenotype_comment_replies
 
+    # get the last 30 updated SNPs for this user
+    @last_updated_snps = @user.snps.order('updated_at desc').limit(30) unless @user.snps.nil?
```
tsujigiri: `@user.snps.nil?` will always be false, since `@user.snps` returns an empty array, if there are no Snps.

philippbayer: So the `unless` block is useless?

philippbayer: Running this on production for User.find(1) takes 23 seconds, ouch

```
Snp Load (23026.7ms)  SELECT  "snps".* FROM "snps" INNER JOIN "user_snps" ON "snps"."name" = "user_snps"."snp_name" INNER JOIN "genotypes" ON "user_snps"."genotype_id" = "genotypes"."id" WHERE "genotypes"."user_id" = $1  ORDER BY plos_updated desc LIMIT 30  [["user_id", 1]]
```

tsujigiri: Joining with `user_snps` is always problematic. Your current approach is probably not much faster, is it? How about we compute the list in the background on a daily basis or so?

philippbayer: It's quite a bit faster :) it runs a whole bunch of sql statements but each only takes ~5ms

```
Benchmark.measure {        @last_30_papers = Set.new
    if !@user.genotypes.nil?
      # get the last 30 updated SNPs for this user
      # we do it the other way around - check the last 20 updated papers for each type, see if their snps are in user's collection
      # the upper if is important since without it even users without SNPs will trigger this
      %w(Snpedia Mendeley GenomeGov Plos).each do |source|
        "#{source}Paper".constantize.last(30).each do |p|
          p.snps.each do |s|
            @last_30_papers.add(p) if s.users.include? @user
          end
        end
      end
      # convert to array, sort by updated_at, get newest update first
      @last_30_papers = @last_30_papers.to_a.sort_by(&:updated_at).reverse
      # in the best case we have 4*30 papers for this user now, in the worst case we have 0
      @last_30_papers = @last_30_papers[0..30]
    end

}
```
on the server with a two different users returns

    => #<Benchmark::Tms:0x00000006857428 @label="", @real=3.512849297374487, @cstime=0.0, @cutime=0.0, @stime=0.04999999999999982, @utime=0.6900000000000004, @total=0.7400000000000002>

This is a run with a randomly chosen user with uploaded genotype - 

With User.find(1) `@real` goes down to 1.6s, probably because it cached stuff from me playing around. 

But having a worker for that table would probably be the most sane way - have a worker that makes a list of 4*30 last updated workers (or have all API workers report their last updated ones? sounds dodgy), so that the user controller can just check whether current_user has any connection to that list.

app/controllers/users_controller.rb
```diff
@@ -112,6 +111,24 @@ def show
     @phenotype_comment_replies.sort! { |b,a| a.created_at <=> b.created_at }
     @paginated_phenotype_replies = @phenotype_comment_replies
 
+    @last_30_papers = Set.new
+    if !@user.genotypes.nil?
+      # get the last 30 updated SNPs for this user
+      # we do it the other way around - check the last 20 updated papers for each type, see if their snps are in user's collection
+      # the upper if is important since without it even users without SNPs will trigger this
+      %w(Snpedia Mendeley GenomeGov Plos).each do |source|
+        "#{source}Paper".constantize.last(30).each do |p|
```
tsujigiri: Just putting the classes themselves into an array is probably slightly more readable.

app/controllers/users_controller.rb
```diff
@@ -112,6 +111,24 @@ def show
...
+          p.snps.each do |s|
+            @last_30_papers.add(p) if s.users.include? @user
```
tsujigiri: Try if `s.users.exists?(@user.id)` is faster. `s.users.include?(@user)` will probably load all the users, `exists?` runs the minimal query necessary to find out if the user can be found, without loading it.

app/controllers/users_controller.rb
```diff
@@ -112,6 +111,24 @@ def show
...
+          end
+        end
+      end
+      # convert to array, sort by updated_at, get newest update first
+      @last_30_papers = @last_30_papers.to_a.sort_by(&:updated_at).reverse
+      # in the best case we have 4*30 papers for this user now, in the worst case we have 0
+      @last_30_papers = @last_30_papers[0..30]
```
tsujigiri: `@last_30_papers.take(30)` is probably more readable.

app/controllers/users_controller.rb
```diff
@@ -112,6 +111,24 @@ def show
...
+    end
+
```
tsujigiri: Putting these kind of things into separate classes helps keeping the controllers small and makes them (the things) more easily unit-testable. Maybe `app/services/last_updated_snps.rb` or `app/services/last_updated_snps_service.rb`, for example. Not sure about the naming. Both have advantages and disadvantages.

tsujigiri: Another suggestion: You can write this in a more compact and functional fashion:

```ruby
[SnpediaPaper, MendeleyPaper, GenomeGovPaper, PlosPaper].flat_map do |papers|
  papers.last(30).select do |paper|
    paper.snps.any? { |s| s.users.exists?(@user.id) }
  end
end.sort_by(&:updated_at).reverse.take(30)
```

app/controllers/users_controller.rb
philippbayer: On the server they both do the same thing:

```
irb(main):009:0> s.users.exists? user        
D, [2016-09-18T11:54:49.401479 #24206] DEBUG -- :   User Exists (3.3ms)  SELECT  1 AS one FROM "users" INNER JOIN "genotypes" ON "users"."id" = "genotypes"."user_id" INNER JOIN "user_snps" ON "genotypes"."id" = "user_snps"."genotype_id" WHERE "user_snps"."snp_name" = $1 AND "users"."id" = $2 LIMIT 1  [["snp_name", "rs472703"], ["id", 5288]]
=> true
irb(main):010:0> s.users.include? user
D, [2016-09-18T11:54:55.606039 #24206] DEBUG -- :   User Exists (2.8ms)  SELECT  1 AS one FROM "users" INNER JOIN "genotypes" ON "users"."id" = "genotypes"."user_id" INNER JOIN "user_snps" ON "genotypes"."id" = "user_snps"."genotype_id" WHERE "user_snps"."snp_name" = $1 AND "users"."id" = $2 LIMIT 1  [["snp_name", "rs472703"], ["id", 5288]]
=> true
```

app/controllers/users_controller.rb
```diff
@@ -112,6 +111,7 @@ def show
     @phenotype_comment_replies.sort! { |b,a| a.created_at <=> b.created_at }
     @paginated_phenotype_replies = @phenotype_comment_replies
 
+    @@last_30_papers = LastUpdatedSnpsService.get_last_thirty_updated_snps
```
philippbayer: This seems to be a buggy comment, the file itself is fixed and correct here in the PR

spec/features/arriving_as_user.rb
```diff
@@ -0,0 +1,28 @@
...
+    end
+
+    scenario 'the user arrives on landing page' do
+      visit root_path
+
+      expect(page).to have_content('Hello Potato!')
+      expect(page).to have_content('rs1234')
+
+      # Test Your last 30 updated SNPs feed
+      expect(page).to have_content('Received new data from Mendeley')
```
gedankenstuecke: shouldn't this read "new data from Snpedia"?

app/services/last_updated_snps_service.rb
```diff
@@ -0,0 +1,9 @@
+class LastUpdatedSnpsService
```
tsujigiri: You can fix all of these running `bundle exec rubocop --auto-correct --only Style/FrozenStringLiteralComment`. You may have to merge master into your branch first.

philippbayer: Thanks! Will run that in a separate PR

spec/features/arriving_as_user.rb
philippbayer: You're right :P

philippbayer:
I guess it can be merged?

PR #333, opened 2016-08-19
OpenSNP front-end redesign

MateusJabour:
@philippbayer @gedankenstuecke


.env.example
```diff
@@ -1,20 +0,0 @@
...
-export REDIS_URL=redis://localhost
-export FITBIT_CONSUMER_KEY=foo
-export FITBIT_CONSUMER_SECRET=bar
-export PLOS_API_KEY=foo
-export RECAPTCHA_PRIVATE_KEY=6LeIxAcTAAAAAGG-vFI1TnRWxMZNFuojJ4WifJWe
-export RECAPTCHA_PUBLIC_KEY=6LeIxAcTAAAAAJcZVRqyHh71UMIEGNQ_MXjiZKhI
-export SIDEKIQ_PASSWORD=foo
-export MENDELEY_CONSUMER_KEY=foo
-export ERRBIT_API_KEY=foo
-export ERRBIT_HOST=localhost
```
tsujigiri: `.env` is in the `.gitignore` file. The idea is that the example file works for testing and to run the app, but you still have the option to change your `.env` without having to take care of not accidentally commit any API keys you don't want to publish.

config/database.yml.example
```diff
@@ -1,21 +0,0 @@
...
-  <<: *base
-  database: snpr_test
-
-production:
-  <<: *base
-  database: snpr_production
-
-staging:
-  <<: *base
-  database: snpr_production
```
tsujigiri: See [comment about the `.env.example` file](https://github.com/openSNP/snpr/pull/333/commits/a5a19eaebb34119b5975463883538c4d964855de#r75593377).

.gitignore
```diff
@@ -33,3 +33,5 @@ vendor/cache
 logfile
 .env
 .DS_Store
+Vagrantfile
+serverscript
```
tsujigiri: Both files are in the repository. Why do you think they should be added here?

MateusJabour: That was a thing that I forgot to change, I will take them off from this file.

.env.example
MateusJabour: So, what should I do about this and the database file?

philippbayer: You can remove both files in a new commit using 'git rm', doing 'git status' to check, and then pushing - everything you push will be added to this pull request.
Then you can also look at the Hound comments, we usually fix those, they don't take long to fix

app/controllers/fitbit_profiles_controller.rb
```diff
@@ -9,7 +9,7 @@ def index
     @fitbit_profiles = FitbitProfile
       .includes(:user)
       .order("#{sort_column} #{sort_direction}")
-      .paginate(page: params[:page], per_page: 20)
+      .paginate(page: params[:page], per_page: 15)
```
MateusJabour: I think this is unnecessary to change, what do you think? @philippbayer @gedankenstuecke @tsujigiri

philippbayer: Hound only comments on what you changed - in this case lines 10 to 12 should be aligned. It's always good to obey the Hound, so that we have a consistent as possible codebase :)

philippbayer:
I just tried to see whether I could fix the PR for you but it doesn't seem to like it - could you please add me as a collaborator in your repository's Settings -> Collaborators & Teams? Then I'll push directly to your branch, that should be easiest 

So far it doesn't look like much changes, about 20 files need some decision on what to keep and what to delete

philippbayer:
Tried to fix it - pushed to the wrong branch first >.<

philippbayer:
Placeholder - working on this today and tomorrow. Hopefully we'll finally have this merged asap

gedankenstuecke:
Has been basically merged through #365. Thanks for all the work @MateusJabour!

PR #334, opened 2016-09-04
[WIP] Linking Phenotypes to SNP

raivivek:
Opening the official PR. Need to polish quite a few things before we can progressively start merging things into master.
- [ ] Complete the worker unit tests
- [ ] Resolve the merge conflicts (merging `structure.sql` seems tricky) - _Probably no need to merge structure.sql. Rather, merge the migration files and run the migration on server to update structure.sql?`_
- [ ] Replace simple string lookup with a RegexExp matcher
- [ ] Regulate database calls for Phenotype table
- [ ] Tweak scoring system

cc/ @philippbayer @gedankenstuecke


app/views/snps/show.html.erb
```diff
@@ -59,6 +63,22 @@
       We don't have enough users yet.
     <%end%>
   </div>
+
+  <div class="span3 columns">
+    <center><h6><a href="#" rel="tooltip" data-html="true" title="Phenotypes possibly associated with this SNP.">Phenotypes</a></h6></center>
+    <% if @snp.phenotypes.count %>
+      <table id="phenotype_table" class="table table-striped">
+        <% PhenotypeSnp.where(:snp_id => @snp.id).order(:score).reverse.each do |p| %>
```
philippbayer: Maybe move this piece into the Phenotype-SNP model or SNPs-show-controller?

Like 

```@ordered_phenotype_snps  = PhenotypeSnp.where(:snp_id => @snp.id).order(:score).reverse```

**Comment by philippbayer** [comment]
:+1: Tests, clean looking code, this looks all very nice!

File: app/views/snps/show.html.erb (hunk 1 above)
Comment (vivekiitkgp) [inline app/views/snps/show.html.erb]: Ah! Makes sense.

File: app/models/snp.rb [hunk 2]
```diff
@@ -43,6 +46,15 @@ def self.update_frequencies
     end
   end
 
+  def self.update_phenotypes
+    max_age = 31.days.ago
+
+    Snp.select([:id, :phenotype_updated]).where(phenotype_updated < max_age)
```
tsujigiri: Shouldn't this be `.where('phenotype_updated < ?', max_age)`? You also don't need the square brackets, `.select(:id, :phenotype_updated)` works as well.

app/models/snp.rb
```diff
@@ -43,6 +46,15 @@ def self.update_frequencies
...
+       .find_each do |snp|
+      Sidekiq::Client.enqueue(LinkSnpPhenotype, snp.id)
```
tsujigiri: `LinkSnpPhenotype.perform_async(snp.id)` is preferable.

app/views/snps/show.html.erb
```diff
@@ -7,7 +7,7 @@
     <table class="table table-striped">
       <tr>
         <td><b>Name</b></td>
-        <td><%=@snp.name%></td>
+        <td><%=@snp.name%> </td>
```
tsujigiri: Is there a reason for the added space? Also, in general, it's better to add a bit of space within the angle brackets to improve readability:
```erb
<td><%= @snp.name %></td>
```

app/models/snp.rb
vivekiitkgp: Right.

app/views/snps/show.html.erb
vivekiitkgp: Seems to have misplaced the space.

app/views/snps/show.html.erb
```diff
@@ -59,6 +63,22 @@
...
     <%end%>
   </div>
+
+  <div class="span3 columns">
+    <center><h6><a href="#" rel="tooltip" data-html="true" title="Phenotypes possibly associated with this SNP.">Phenotypes</a></h6></center>
+    <% if @snp.phenotypes.count %>
+      <table id="phenotype_table" class="table table-striped">
+        <% @ordered_phenotype_snps.each do |p| %>
+          <tr>
+            <td> <%= Phenotype.find(p.phenotype_id).characteristic %> </td>
```
tsujigiri: This sends a query to the database for every iteration of the loop. You can get the phenotypes all at once, by using `includes` on `PhenotypeSnp`:

```rb
@ordered_phenotype_snps = @snp.phenotype_snps.includes(:phenotype).order(score: :desc)
@ordered_phenotype_snps.each do |phenotype_snp|
  phenotype_snp.phenotype.characteristic
end
```

app/workers/link_snp_phenotype.rb
```diff
@@ -0,0 +1,68 @@
+class LinkSnpPhenotype
+  include Sidekiq::Worker
+  sidekiq_options :queue => :link_snp_phenotype, :unique => true
+  attr_reader :snp, :characteristics, :papers_count
+
+  def perform(snp_id)
+    @snp = Snp.find(snp_id)
+    @characteristics = Phenotype.all.map { |x| x.characteristic }
+    @papers_count = @snp.snp_references.length
```
tsujigiri: `count` is probably better than `length` here. It sends a `COUNT(*)` query to the database. Unless `length` has been overwritten in `ActiveRecord`, which I'm not sure it is, `length` will lead to all the records being loaded just to count them.

app/models/snp.rb
vivekiitkgp: Ah. Okay. That's how I was running the test job in console. Just curious about the preference? Both are performed asynchronously anyway, are they not? @tsujigiri

app/workers/link_snp_phenotype.rb
```diff
@@ -0,0 +1,68 @@
...
+  # metadata. The weight for each count is received from the calling function.
+  def score_paper(paper_type, weight)
+    score = {}
+
+    papers = @snp.public_send(paper_type)
+
+    # search for each phenotype one by one in the papers' metadata
+    @characteristics.each do |chr|
+      papers.each do |p|
+        title = if p.class.method_defined? :title then p.title else p.summary end
```
tsujigiri: How about this instead:
```ruby
p.respond_to?(:title) ? p.title : p.summary
```

db/migrate/20160515212622_create_join_table_phenotype_snp.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.float :score, :default => 0
+      t.timestamps
+    end
+
+    add_index :phenotype_snps, [:snp_id, :phenotype_id], :unique => true
+  end
+
+  def self.down
+    drop_table :phenotype_snps
+  end
```
tsujigiri: If you rename `self.up` to `change` and remove the `self.down` method, Rails can infer the corresponding `drop_table` from the `create_table` call.

app/models/snp.rb
tsujigiri: Yes, they are. I'd prefer `perform_async` because I find it more readable, as it's less boilerplate code.

app/views/snps/show.html.erb
tsujigiri: Here is the documentation, if you are interested:
http://api.rubyonrails.org/classes/ActiveRecord/QueryMethods.html#method-i-includes

tsujigiri:
Shouldn't Hound complain about the hash-rocket syntax? (Prefer `{ foo: :bar }` over `{ :foo => :bar }`) I think it should. :stuck_out_tongue:

PR #336, opened 2016-11-16
added survey-email feature

gedankenstuecke:
In order to re-send the survey emails for crowdAI, this time from survey@opensnp.org we need

1. a way to set up the new email address alongside the other one
2. a way to only send to people who haven't answered that survey already
3 a way to exclude people who have no genetic data uploaded

I implemented 1. by adding the specific settings to the `.env` and created a `survey` mailer-definition that uses these `ENV` settings (see .env.example). 

For 2. and 3. I added a new rake task `survey:send` that can be run with or without specifying a list of users that should not get the survey emails. 

To run the task solitarily just run `rake survey:send`. To exclude a list of users you can run `rake survey:send EXCLUDEFILE=/path/to/file.txt` with `file.txt`containing one user-ID per line. 

Do you think that makes sense?

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -19,7 +19,11 @@ export MENDELEY_CONSUMER_KEY=foo
 export ERRBIT_API_KEY=foo
 export ERRBIT_HOST=localhost
 
+# username for smtp
 export SURVEY_EMAIL_USER=foobar
+# pw for smtp
 export SURVEY_EMAIL_PASSWORD=foobar
+# server-address for smtp
 export SURVEY_EMAIL_ADDRESS=foobar
+# port for smtp
 export SURVEY_EMAIL_PORT=foobar
--- a/.env.example
+++ b/.env.example
@@ -18,3 +18,8 @@ export SIDEKIQ_PASSWORD=foo
 export MENDELEY_CONSUMER_KEY=foo
 export ERRBIT_API_KEY=foo
 export ERRBIT_HOST=localhost
+
+export SURVEY_EMAIL_USER=foobar
+export SURVEY_EMAIL_PASSWORD=foobar
+export SURVEY_EMAIL_ADDRESS=foobar
+export SURVEY_EMAIL_PORT=foobar
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -76,10 +76,10 @@ default :from => "donotreply@opensnp.org"
 
 	def survey(user)
 		@user = user
-		delivery_options = { user_name: ENV.fetch('SURVEY_EMAIL_USER'),
-                         password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
-                         address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
-												 port: ENV.fetch('SURVEY_EMAIL_PORT') }
+		delivery_options = { 	user_name: ENV.fetch('SURVEY_EMAIL_USER'),
+													password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
+													address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
+													port: ENV.fetch('SURVEY_EMAIL_PORT') }
 		mail(:subject => "openSNP: A one-question survey on height", :to => @user.email,:from => "survey@opensnp.org",delivery_method_options: delivery_options)
 	end
 
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -74,6 +74,15 @@ default :from => "donotreply@opensnp.org"
     mail(:subject => "openSNP meets crowdAI: Crowdsourcing the Prediction of Phenotypes from Genotypes", :to => @user.email)
   end
 
+	def survey(user)
+		@user = user
+		delivery_options = { user_name: ENV.fetch('SURVEY_EMAIL_USER'),
+                         password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
+                         address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
+												 port: ENV.fetch('SURVEY_EMAIL_PORT') }
+		mail(:subject => "openSNP: A one-question survey on height", :to => @user.email,:from => "survey@opensnp.org")
+	end
+
   def dump(target_address, link)
     @link = link
     mail(:subject => "openSNP.org: The data dump you requested is ready to be downloaded",:to => target_address)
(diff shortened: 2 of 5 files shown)


.env.example
```diff
@@ -18,3 +18,8 @@ export SIDEKIQ_PASSWORD=foo
 export MENDELEY_CONSUMER_KEY=foo
 export ERRBIT_API_KEY=foo
 export ERRBIT_HOST=localhost
+
+export SURVEY_EMAIL_USER=foobar
+export SURVEY_EMAIL_PASSWORD=foobar
+export SURVEY_EMAIL_ADDRESS=foobar
+export SURVEY_EMAIL_PORT=foobar
```
tsujigiri: I don't understand what these are for. Isn't it enough to just set the sender's email address?

lib/tasks/send_survey.rake
```diff
@@ -0,0 +1,27 @@
+namespace :survey do
+  desc "send survey"
+  task :send => :environment do
+    # invoke script with rake survey:send EXCLUDEFILE=/path/to/file.txt
+    # read list of users to exclude
+    # should contain one user-ID per line
+    if ENV['EXCLUDEFILE']
+      exclude_users = File.readlines(ENV['EXCLUDEFILE']).each {|l| l.chomp!}
```
tsujigiri: You might want to use `map` instead of `each`. `each` just iterates over the elements of the Array and returns `nil`.

.env.example
gedankenstuecke: But aren't we then sending our email through our own relay instead of through the gmail smtp? The one that is having trouble with being blocked by gmail right now? 😉

tsujigiri: Ah, I thought it was only about the sender address. Are you sure we still have that spam issue, even though our score looks pretty good? If so, I guess we need to go through GMail, then.

lib/tasks/send_survey.rake
```diff
@@ -0,0 +1,27 @@
...
+    # send survey to each user that
+    # a) allows us emailing them
+    # b) has genetic data
+    # c) is not already in the DB
+    User.where(:message_on_newsletter => true).find_each do |u|
+      unless exclude_users.include?u.id.to_s
+        if u.genotypes.length > 0
+          UserMailer.survey(u).deliver_now
+          # wait for one minute so we don't crash the google mail daily limit
+          sleep(1.minutes)
```
gedankenstuecke: done

lib/tasks/send_survey.rake
gedankenstuecke: done

lib/tasks/send_survey.rake
```diff
@@ -0,0 +1,27 @@
+namespace :survey do
+  desc "send survey"
```
gedankenstuecke: done

app/mailers/user_mailer.rb
```diff
@@ -74,6 +74,15 @@ def newsletter(user)
     mail(:subject => "openSNP meets crowdAI: Crowdsourcing the Prediction of Phenotypes from Genotypes", :to => @user.email)
   end
 
+	def survey(user)
+		@user = user
+		delivery_options = { user_name: ENV.fetch('SURVEY_EMAIL_USER'),
```
gedankenstuecke: whops, good catch 😂

lib/tasks/send_survey.rake
```diff
@@ -0,0 +1,27 @@
+namespace :survey do
+  desc 'send survey'
+  task :send => :environment do
+    # invoke script with rake survey:send EXCLUDEFILE=/path/to/file.txt
+    # read list of users to exclude
+    # should contain one user-ID per line
+    if ENV['EXCLUDEFILE']
+      exclude_users = File.readlines(ENV['EXCLUDEFILE']).map { |l| l.chomp! }
```
tsujigiri: What the talking dog wants is this:

```ruby
.map(&:chomp!)
```
In case you were wondering.

gedankenstuecke:
gotta run to my plane to LAX, will do more review later!

tsujigiri:
Safe travels! :airplane:

gedankenstuecke:
Thanks! Would be awesome if one or you folks could give it a try. Worked on my end so far 💖😂

Talk later! ✈️

--
Sent from a mobile device
mobile: +49 176 213 044 66

While I may be sending this email outside my normal office hours, I have no expectation to receive a reply outside yours

lib/tasks/send_survey.rake
```diff
@@ -0,0 +1,27 @@
+namespace :survey do
```
gedankenstuecke: what does it want here?

philippbayer: I think it wants you to add '# frozen_string_literal: true`' (or false) to the top of the file. I'm not sure whether we need it. This will make strings immutable, and is default in Ruby 3, it just wants you to be explicit about whether you want mutability, I guess?

gedankenstuecke: mh, maybe. ignored for now :D

gedankenstuecke:
Okay, I think the @houndci-bot and @tsujigiri are more pleased now. Would love a second round of review now (sent from the most crappy Motel WiFi ever, it makes you think you could work, alas: no!)

philippbayer:
The only thing I would complain about is that the next time we run a survey we'll have to replace that text again and we'll definitely forget to change the mail header :) But I think that's a task for wider refactoring as we have the same problem in other mailers, too.

lib/tasks/send_survey.rake
tsujigiri: Well, I would have said just add it. But since it's already merged...

gedankenstuecke: But then we should add it for all, right?

app/mailers/user_mailer.rb
```diff
@@ -74,6 +74,15 @@ def newsletter(user)
...
 
+	def survey(user)
+		@user = user
+		delivery_options = { 	user_name: ENV.fetch('SURVEY_EMAIL_USER'),
+													password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
+													address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
+													port: ENV.fetch('SURVEY_EMAIL_PORT') }
+		mail(:subject => "openSNP: A one-question survey on height", :to => @user.email,:from => "survey@opensnp.org",delivery_method_options: delivery_options)
+	end
+
```
gedankenstuecke: that's weird, because the hound complained before and now it doesn't o_O

tsujigiri: Maybe it only complains once. :wink:

lib/tasks/send_survey.rake
tsujigiri: Hm... same here. Strange. :thinking:

gedankenstuecke: Phew, happy to hear that it's not again me doing something wrong :D

Issue #176, opened 2015-06-04
Bug in Variations you did not enter


duartemolha:
Hi
when in the section "Variations you did not enter", when you initially click on one (in thie case Astigmatism) I am greeted with a popup where you can either write a text to describe your variation or choose from the list below.

I initially clicked one of the options and the free text option on top disappeared. However, after reading the options I decided none of them really described my phenotype correctly so because I wanted to use the free text option I closed the popup without clicking OK

When I clicked on Astigmatism option again the free-text option would not show up again and one the multiple choice options was selected.

I believe if the user does not click ok and navigates out of the popup, the multiple choice he innitialy clicked should not be accepted as valid and thus when navigating back to the variation, the free-text option should be visible.

tsujigiri:
Thank you very much for reporting this! You earned +1 bug reporting karma. :) One of us will look into it as soon as he finds the time.

duartemolha:
you're welcome.

gedankenstuecke:
This bug even goes a bit further, because it removes **all** free-text input fields as I've just found out. This is due to the fact that we do a blanket `$('input[type=radio]').click(function()` in the view, without specifying to what the button should belonging to. 

So fixing that might be a bit more complicated?

7yl4r:
I think this code snippet is all that is needed, but I'm not sure what file it should go in (I'm unfamiliar with ruby project layout). 

```javascript
$('.close').click(function(){
    $('input[type=text]').show();
})
```

gedankenstuecke:
I guess that could go into [`app/views/user_phenotypes/_form.html.erb`](https://github.com/openSNP/snpr/blob/67ff132c3b97a9c626d4a4c07a520c8e37ed3d7f/app/views/user_phenotypes/_form.html.erb)? But I assume that @tsujigiri has a better idea of how to correctly do it 😉

gedankenstuecke:
Ah, thanks. Do we already have a Issue open for that?

tsujigiri:
Not that I know of.

7yl4r:
[...]

Issue #329, opened 2016-08-10
Remove Pagination


gedankenstuecke:
[Following this blog post](https://blog.jooq.org/2016/08/10/why-most-programmers-get-pagination-wrong/) @tsujigiri and I briefly discussed this on _Gitter_: 

The SNP index (and I think all the other index pages as well) doesn't actually need the pagination where you can jump to a given page of SNPs (currently there are 236591 pages…). No one uses them. Instead just offering the ⬅️  ➡️  buttons would be good enough. 

This should a) remove clutter from the index pages and b) make things faster.

7yl4r:
This should probably be implemented as a feature of [will_paginate](https://github.com/mislav/will_paginate) (or finding an alternative), rather than rolling a custom solution for this project.

gedankenstuecke:
Yep, fully agreed. We should figure out whether there's such a thing that we could just replace our current solution with. 😃

audy:
[ajax-datatables-rails](https://github.com/antillas21/ajax-datatables-rails) is pretty good. I use it for tables with 100,000s of rows. And you get free search.

tsujigiri:
Looks like we only need to add a `page_links: false` to the `will_paginate` call.

https://github.com/mislav/will_paginate/wiki/API-documentation#will_paginatecollection-options

gedankenstuecke:
That sounds good, maybe @7yl4r wants to give it a try, would make a fun first commit 😄

7yl4r:
I was hoping to get a pull request submitted quickly this morning, but I have had some difficulty getting the project set up. I thought I had everything working (looks fine when loaded on localhost), but `bundle exec rake` fails 52/52 tests. :dizzy_face:

I will get it figured out... eventually.

PR #337, opened 2016-12-26
Resolves issues #176 and #329

7yl4r:
I'm new to all of these backend technologies, but these frontend fixes are a nice simple start for me.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -13,13 +13,13 @@
 //= require jquery_nested_form
 //= require_self
 
-$(document).ready(function() { 
+$(document).ready(function() {
     $('#update_user').click(function() {
         $('#update_user').val("Updating...").addClass("disabled");
         setTimeout(function(){ $('#update_user').val("Update Information").removeClass("disabled");},1000);
     });
 
-    //$("#PaperMendeley").tablesorter({sortList: [[2,1]]}); 
+    //$("#PaperMendeley").tablesorter({sortList: [[2,1]]});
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/application.js
```diff
@@ -55,4 +55,9 @@ $(document).ready(function() {
     var $li = $(this).parent("li").toggleClass('open');
     return false;
     });
+
+    // re-show custom text option that may have been previously hidden
+    $('.close').click(function(){
```
tsujigiri: Should the selector be more specific? It's a fix for one specific place, right?

app/helpers/application_helper.rb
```diff
@@ -27,7 +27,7 @@ def previous_or_next_page(page, text, classname)
   end
 
   def page_navigation_links(pages)
-    will_paginate(pages, class: 'pagination', inner_window: 2, outer_window: 0, renderer: BootstrapLinkRenderer, previous_label: '&larr;'.html_safe, next_label: '&rarr;'.html_safe)
+    will_paginate(pages, class: 'pagination', inner_window: 2, outer_window: 0, renderer: BootstrapLinkRenderer, previous_label: '&larr;'.html_safe, next_label: '&rarr;'.html_safe, page_links: false)
```
tsujigiri: It's better to put unrelated changes into individual pull requests. Makes it easier when looking through the history to figure out what was done why.

7yl4r: Thanks for both suggestions. I'll open a new PR for will_paginate and will a second one with a more specific selector string for the other issue.

PR #338, opened 2016-12-26
Update gems

tsujigiri:
* Passwords now need to be 8 characters long (new default in authlogic).
* To login to the Sidekiq web interface, log in to opensnp.org, if you
  have the `admin` flag set, you will find it at the usual place.
* I finally found out why Rubocop didn't complain about the frozen string literal pragma missing and added all of them.
* Rubocop complained about keyword arguments, which is only relevant for Rails 5, so I disabled the cop.
* I removed the `inherit_from: .rubocop_todo.yml` from `.rubocop.yml`, so Hound does complain about **all** the issues. I also added a few tasks for running Rubocop.


tsujigiri:
This should be good to go, now.

gedankenstuecke:
Looks good to me. @philippbayer any complains from your side? 😺

tsujigiri:
Sooo... should I merge or do we wait for @philippbayer to respond? 🙂

gedankenstuecke:
Fire away, I think @philippbayer just arrived in 🇩🇪  this morning from 🇯🇵 , so it would be a few days before he's able to read straight again. 😂

PR #340, opened 2016-12-28
Removes pagination (#329)

7yl4r:

PR #346, opened 2017-02-01
Closes #329 and #340 by amending tylar's PR

### LINKED ISSUE #329: Remove Pagination

gedankenstuecke:


philippbayer:
There are some age-old helpers for this in app/helpers/application_helper.rb which I put there in *2012* to make the pagination look nicer, the methods page_number, previous_or_next_page, and gap, I think these can be deleted? Or are they called somewhere else?

app/helpers/application_helper.rb
```diff
@@ -9,25 +9,8 @@ def sortable(column, title = nil)
   class BootstrapLinkRenderer < ::WillPaginate::ActionView::LinkRenderer
```
philippbayer: These two lines can go too - see line 27, there's a double end there

currently it tries to mash page_navigation_links into Bootstrap which breaks

PR #350, opened 2017-04-03
Adds disclaimer about faked data

philippbayer:
see text


philippbayer:
OK I renamed it to 'Account housekeeping' and added your sentence

PR #362, opened 2017-05-31
Possible fix for CSS missing in Sidekiq web form

philippbayer:
We're currently (and have been for quite a while) missing the CSS and JS from the Sidekiq webform, but only in production.

Their FAQ says to just replace the production.rb sendfile_header: https://github.com/mperham/sidekiq/wiki/Problems-and-Troubleshooting#sidekiq-web-does-not-render-correctly-in-production-but-works-fine-in-development

This implements this, who knows whether it'll work, can't test for obvious reasons

### DIFF
--- a/config/environments/production.rb
+++ b/config/environments/production.rb
@@ -12,10 +12,11 @@ Snpr::Application.configure do
   config.action_controller.perform_caching = true
 
   # Specifies the header that your server uses for sending files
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
at least the hound seems happy now. Will merge as soon as travis approves as well.

PR #364, opened 2017-06-01
Added steps to initialize database

bverpaalen:
Postgres can be a bit of a struggle if you are not used to it so I added extra steps for creating a user and how to give it superuser rights

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -78,7 +78,7 @@ sudo -u postgres psql postgres
 ```
 Make your user a super user
 ```
-ALTER USER myuser WITH SUPERUSER;
+ALTER USER username WITH SUPERUSER;
 ```
 Setup the database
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
That's really useful, thanks for that. Are these platform-dependent steps or should it work for any Unixy-system? Otherwise would be cool to at least give an idea on which system this works. 😊

bverpaalen:
I ran them on Fedora 25. It should work for every linux distribution but I didn't test it

bverpaalen:
Looks good. @ascheben is trying to install it on windows but it's quit the challenge.

gedankenstuecke:
@philippbayer told me about that! We can merge that in later, when/if @ascheben succeeds (I can imagine it being a pain under Windows!)

Issue #375, opened 2017-06-02
(UI) bugs due to frontend overhaul


gedankenstuecke:
With merging #365 we got some (mainly UI) bugs that we need to tend to:

- [x] Glyphicons in production are broken.
- [x] Avatar for @gedankenstuecke doesn't load on *About Us*
- [x] Modals for entering new phenotypes are white text on white background
- [x] Numbered Pagination came back, see #329 for fix
- [x] "Latest Data"-tab loads forever, ineffective database queries?
- [x] Pie charts for allele/genotype frequencies don't have legend.
- [x] Fitbit graphs are also strange.
- [x] Remove #{@user.name} SNPs on _user_not_user.html.erb
- [x] Remove index-column on index.html.erb for users & genotypes
- [x] Phenotype entry autocompletion is somewhat broken (on modal an phenotype page directly)

Add your own things below 😂

gedankenstuecke:
Replacing my avatar with a PNG instead of JPEG worked. No clue why, should investigate this.

gedankenstuecke:
#379 also closes the broken modals. My bad for not switching branches in time.

gedankenstuecke:
With the UI update #223 seems to be at least a bit better now. The Autocomplete on the modal phenotype entry works if you cycle through it with your arrow keys, but it doesn't show the actual selection in a sensible way so far. Might be some CSS/JS updates that killed it.

philippbayer:
Glyphicons have been fixed, see https://github.com/openSNP/snpr/pull/383

gedankenstuecke:
Yay, thanks! 🙏

gvd:
Regarding the phenotype autocomplete. I see the following code:
    $(function () {
        $("#phenotype_characteristic").autocomplete({
            source: ["Retrognathia (Marfan Syndrome)","Eye pigmentation ","Vegetarianism/Preference for Meat","Form of foot","Eye color","A+","MSG tastes...", "....."]
        });

Which is missing some brackets. It should be:
    $(function () {
        $("#phenotype_characteristic").autocomplete({
            source: ["Retrognathia (Marfan Syndrome)","Eye pigmentation ","Vegetarianism/Preference for Meat","Form of foot","Eye color","A+","MSG tastes...", "....."] }); <----------------
        });

gedankenstuecke:
#411 closes the last one of the bugs reported here so far. 🎉

PR #379, opened 2017-06-02
pagination w/o page numbers again

gedankenstuecke:
Another one for #375, this closes the issue of the pagination having page numbers again (bad idea w/ 2+ million SNPs). 

* Mateus changed our own helper to the `will_paginate_bootstrap` gem, this is why our fixed helper didn't work any longer. 
* removed our old helper
* wrote new helper_function that passes `page_links: false` and replaced `will_paginate` calls with helper-calls


gedankenstuecke:
Whops, accidentally pushed the modal fix as well. In any case: The modal views now work again, the text is not white any longer.

PR #387, opened 2017-06-05
Testwide code coverage fix

philippbayer:
This could fix the code coverage test coverage badge, but who knows

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -2,7 +2,7 @@
 ENV['RAILS_ENV'] = 'test'
 unless ENV['CI']
   require 'simplecov'
-  SimpleCov.start('rails')
+  SimpleCov.start
 end
 require File.expand_path('../../config/environment', __FILE__)
 require "test/unit"
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,8 @@ addons:
     packages:
     - libhiredis-dev
   postgresql: '9.3'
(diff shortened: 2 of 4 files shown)


tsujigiri:
The Codeclimate token is already configured as an environment variable over at https://travis-ci.org/openSNP/snpr/settings. Things like this don't really belong anywhere where everyone can see them. Not that I expect anyone to send them false coverage reports. :wink: Still...

test/test_helper.rb
```diff
@@ -2,7 +2,7 @@
 ENV['RAILS_ENV'] = 'test'
 unless ENV['CI']
   require 'simplecov'
-  SimpleCov.start('rails')
+  SimpleCov.start
```
tsujigiri: I'm not sure this is what we want. 😬

gedankenstuecke: meaning?

gedankenstuecke:
> The Codeclimate token is already configured as an environment variable over at https://travis-ci.org/openSNP/snpr/settings. Things like this don't really belong anywhere where everyone can see them. Not that I expect anyone to send them false coverage reports

Ah, @philippbayer and I were following *code climate*'s own how-to and they recommend doing it like this. But of course you're right about it not being nice. I regenerated a new token, put this one in the *Travis CI* backend and deleted the old one from the `.travis.yml`, still seems to work now. :+1:

PR #389, opened 2017-06-06
Put back Rails specific config for SimpleCov

tsujigiri:
Without the `'rails'`, coverage drops to ~15% for the legacy tests, which probably means, it doesn't find all the tests.


gedankenstuecke:
One of the tests took to long on the `travis-ci/push` commit. Just rerunning it now to see whether it passes now 😂

PR #393, opened 2017-06-07
Update ROADMAP.md

gedankenstuecke:
removed done item from ROADMAP. Also wanna see whether the reviewing thing works ;-)


gedankenstuecke:
Thanks, I'd kind of advocate for removing the hard barrier of not allowing to merge PRs w/o review. Just protecting `master` from direct pushing should be good enough. 

Also: I think the only 3 people with permissions to merge right now are @philippbayer @tsujigiri and me. And we three also have the rights to remove this setting that requires a review, so it's a bit moot.

Opinions on that @tsujigiri?

tsujigiri:
Like I said, in general and with a team as small as ours, I think it's enough to have this as a convention, rather than enforced by Github. I don't mind whether we remove it or keep it.

gedankenstuecke:
Ok, I'll remove it, seems like it's just a fake-ish roadblock to merging things quickly when the need arises (breaking production 😉 )

PR #394, opened 2017-06-07
[MRG] fixed message deleting + added tests, closes #391

### LINKED ISSUE #391: Deleting messages doesn't work
I'm getting an unauthorized when trying to delete a message on the production system.

gedankenstuecke:
With this #391 should be closed as you can delete messages you've received once again. It also adds some tests to the messages.

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -48,11 +48,10 @@ class MessagesController < ApplicationController
     if message
       message.destroy
       flash[:notice] = 'Message deleted'
-      redirect_to '/users/' + current_user.id.to_s + '#messages'
     else
       flash[:warning] = 'Something went wrong!'
-      redirect_to '/users/' + current_user.id.to_s + '#messages'
     end
+    redirect_to '/users/' + current_user.id.to_s + '#messages'
   end
 
   private
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -44,13 +44,14 @@ class MessagesController < ApplicationController
   end
 
   def destroy
-    message = Message.where(from_id: current_user.id, id: params[:id]).first
+    message = Message.find_by(id: params[:id])
     if message
       message.destroy
       flash[:notice] = 'Message deleted'
       redirect_to '/users/' + current_user.id.to_s + '#messages'
     else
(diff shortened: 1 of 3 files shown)


gedankenstuecke:
All robot reviewers are fine, test coverage increases, bug is fixed. :shipit:

tsujigiri:
Ah, I missed [the line in `require_owner`](https://twitter.com/HoracioMelvin/status/872222085657112577). Hm... whose `user_id` is this? The sender's?

tsujigiri:
Who should be able to delete messages and should they still be visible to the other correspondent? Or is it enough for the sender to be able to delete messages, which then disappear for both?

tsujigiri:
Also: `from_id` should be the same as `user_id`. `user_id` is checked in `require_user`. So, if I understand this correctly, this shouldn't even change anything. :question:

gedankenstuecke:
I think this needs a not-so-short description:

The `user_id` gives the user who owns the message. Our message system duplicates each message when it's sent. So Alice(`user_id=1`) creates a message to Bob (`user_id=2`). What happens is that the message from Alice looks like this: 

`id: 1, user_id: 1, from_id: 1, to_id: 2, subject: "foo", body: "bar", sent: true`. 

The `message.rb` model duplicates this so that Bob owns a copy too: 

`id 2, user_id: 2, from_id: 1, to_id: 2, subject: "foo", body: "bar", sent: false`

Now each of them is capable of deleting their copy of the message from their respective in/outbox while the other one remains a copy of it. That's why we just need to check for ownership through `user_id` instead of the `to/from` field. And that's also the reason why `from_id != user_id` for received messages unless you message yourself. 

Does that help? 😄

tsujigiri:
Ah, I wasn't aware of the duplicating part. Neat. :slightly_smiling_face:

gedankenstuecke:
Oh, and see also `spec/features/messaging_spec.rb` that now also tests for 

1. deleting messages you received from other users
2. you can't access other user messages 

There's no explicit test for not deleting, but both are handled by `require_owner`, so guess that's okay for now?

tsujigiri:
It's an improvement already. :wink:

gedankenstuecke:
I was too tired yesterday night to figure out how to make the appropriate post-request via the tests. 😂

PR #400, opened 2017-06-10
Use new Codeclimate test reporter

### LINKED ISSUE #398: Test Coverage Calculation
I'm a bit confused when it comes to that:

Codeclimate says that all of the json part in the `snps_controller.rb` [would not be covered](https://codeclimate.com/github/openSNP/snpr/coverage/59390933f22b7500010ac267). At the same time there's at least some testing going on in [`spec/requests/snps_spec.rb`](https://github.com/openSNP/snpr/blob/master/spec/requests/snps_spec.rb). What am I missing? 😂

tsujigiri:
It let's us combine the reports from both, rspec and test/unit.

Fixes #398

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,9 @@ before_script:
   - cp config/database.yml.ci config/database.yml
   - cp .env.example .env
   - bundle exec rake db:setup
+  - curl -L https://codeclimate.com/downloads/test-reporter/test-reporter-latest-linux-amd64 > bin/cc-test-reporter
+  - chmod +x bin/cc-test-reporter
+  - bin/cc-test-reporter before-build
 script: bundle exec rake
 deploy:
   provider: script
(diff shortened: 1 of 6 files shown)


tsujigiri:
"codeclimate/coverage — 50.42% (-27.7%)" :grimacing:

gedankenstuecke:
Ouch!

tsujigiri:
I cannot really tell if this worked or not, as the coverage of individual files onlz seems to show for the master branch on Codeclimate. If I interpret that correctly.

tsujigiri:
At least the commands, that I used to combine and report everything didn't complain.

tsujigiri:
Logically, I guess this makes sense though, since 50.42% lies between RSpec's 41.23% and test/unit's 19.65%, accounting for some overlap. I'm not sure where Codeclimate got it's previous coverage from, though.

tsujigiri:
[RSpec wasn't configured for Rails, before.](https://github.com/openSNP/snpr/pull/400/files#diff-93830fa29d616f7c87903d08b5b1b29aL3) That may be where the difference comes from. 

@gedankenstuecke You may or may not want to reconsider your 80% goal. :grimacing:

tsujigiri:
I'll change the SimpleCov config, so we see if my assumption is correct.

tsujigiri:
"codeclimate/coverage — 63.5% (-14.6%)"

I'd go with SimpleCov configured for Rails, as I'd assume that it's more accurate, even if it means the coverage drops. What do you think?

gedankenstuecke:
👍 guess that's a good reason to write more tests for me. 😂

PR #403, opened 2017-06-15
Adding Newsletter: Redesign + Diego Gómez

gedankenstuecke:
Also fixes a duplication in our footer in the `user_mailer.html.erb` layout.

### DIFF
--- a/app/mailers/user_mailer.rb
+++ b/app/mailers/user_mailer.rb
@@ -97,7 +97,7 @@ class UserMailer < ActionMailer::Base
                          password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
                          address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
                          port: ENV.fetch('SURVEY_EMAIL_PORT') }
-    mail(subject: 'openSNP: Read our survey results and meet Open Humans',
+    mail(subject: 'openSNP: a fresh design and a plea for help',
          to: @user.email,
          from: 'survey@opensnp.org',
          delivery_method_options: delivery_options)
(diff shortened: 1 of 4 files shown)


app/mailers/user_mailer.rb
```diff
@@ -97,7 +97,7 @@ def survey(user)
                          password: ENV.fetch('SURVEY_EMAIL_PASSWORD'),
                          address: ENV.fetch('SURVEY_EMAIL_ADDRESS'),
                          port: ENV.fetch('SURVEY_EMAIL_PORT') }
-    mail(subject: 'openSNP: Read our survey results and meet Open Humans',
+    mail(subject: 'openSNP: a fresh design and a plea for help',
```
tsujigiri: Maybe with a capital A after the colon? :woman_teacher:

tsujigiri:
We really need a nicer way to send these.

gedankenstuecke:
It's already much nicer since we got the layout for the mailer but yep, it's not really fun like this. But what would be a good way to make it easier?

tsujigiri:
It could be as simple as a form, the content of which we pass into [`simple_format`](http://api.rubyonrails.org/classes/ActionView/Helpers/TextHelper.html#method-i-simple_format) and send it off.

gedankenstuecke:
Good idea, for that we'd need some flag which users are allowed to send out newsletters I guess  😉

tsujigiri:
I'd just hard-code a list somewhere. It's not like this changes very often.

tsujigiri:
Or maybe an environment variable to avoid putting email addresses on here, where they are visible to spam bots.

gedankenstuecke:
Yep, that sounds good to me. 😄

PR #409, opened 2017-06-19
Fix phenotype entry autocomplete for #375

gvd:
It was missing a few brackets

### DIFF
--- a/app/views/phenotypes/new.html.erb
+++ b/app/views/phenotypes/new.html.erb
@@ -9,6 +9,7 @@
         $("#phenotype_characteristic").autocomplete({
           source: <%= raw(phenotype_array.to_json) %>
         });
+      });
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
Thanks so much for the contribution @gvd (please email me your address to bgreshake@gmail.com if you want to get some openSNP stickers as a small sign of appreciation)! 

It seems to me that this is not strictly speaking the issue of #375, it's about the modal a user is seeing when being logged in, e.g. `https://opensnp.org/users/1`. There's a list of phenotypes where users can click `Enter Phenotype` and it opens a modal window, that should allow for the autocomplete as well. The code for that is in `app/views/user_phenotypes/_new.html.erb`. Maybe you have an idea what's not working there as well? :)

PR #411, opened 2017-06-19
fixed modal autocomplete! closes #375

### LINKED ISSUE #375: (UI) bugs due to frontend overhaul

gedankenstuecke:
This was a tough one! The issue was the CSS, not the JS! The `z-index` of the autocomplete CSS was too low and basically behind the modal. 

So technically it would all work, it just happened to be hidden by the modal. Can't tell you how long I spent with this bug and trying to figure out how to solve it! Hard to debug if you don't even get an error in your JS-console. 😂 😢 

This also solves what #409 tried to fix!

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -4,3 +4,6 @@
   *= require_self
   *= require style
  */
+ .ui-autocomplete{
+     z-index:1050;
+ }


philippbayer:
one line fix, very nice work

PR #413, opened 2017-06-20
Fix modal disappearance, closes #176

### LINKED ISSUE #176: Bug in Variations you did not enter

gedankenstuecke:
This is another super old bug that's been around for a while. 

The issue was (#176): If a user opens a phenotype-entry modal on their dashboard, checks one of the radio buttons and closes it, the text entry field will disappear and can't be recovered unless they're doing a hard reload of the page.

What happens after this fix, when a user closes the modal: 
* The hidden elements are shown again
* the selected checkbox will be unchecked. 

This is more how the modal should behave after re-opening.

### DIFF
--- a/app/views/layouts/_modal.html.erb
+++ b/app/views/layouts/_modal.html.erb
@@ -2,7 +2,7 @@
   <div class="modal-dialog" role="document">
     <div class="modal-content">
       <div class="modal-header">
-        <button type="button" class="close" data-dismiss="modal">×</button>
+        <button id="modal_close_<%= id %>" type="button" class="close" data-dismiss="modal">×</button>
         <%= content_for id.to_sym%>
       </div>
       <div class="modal-body">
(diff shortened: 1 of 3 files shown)


philippbayer:
Do we need an MT-M fixer in the parsing job as well? Right now only the rake tasks fixes everything, but that will only run once every few weeks

gedankenstuecke:
This in principle only applies for new SNPs, right? Because already existing ones will not be overwritten?

philippbayer:
Yes

gedankenstuecke:
I'd say it's okay to keep it in like this for now, but can also adjust the parser if you think that's nicer :)

Issue #418, opened 2017-07-10
Remove user SNPs tab?


philippbayer:
I was thinking about speedups and realised something:

https://opensnp.org/snps/rs12979860

Does anybody even use the user-alleles tab on that page? People are maybe interested in their own allele and the summary stats, but I cannot see a case where you'd like to scroll through 1498 alleles. Can we just remove that tab and ease the load on the server from all the crawling bots?

gedankenstuecke:
At least one user, Ian, always complains when those stats are not up to date actually, he's using them to look into super rare alleles and to find who has those. Maybe there's a way to make this specific use case easier?

philippbayer:
We could also put those into a second view, so these wouldn't need to be loaded for every crawler?

gedankenstuecke:
Isn't the tab already dynamically loaded only when people click on it? Or was this a former try that we didn't end up using? :D

gedankenstuecke:
Nope, we have the ajax thing for that in place, e.g. 

```
<script type="text/javascript">
  $.ajax({
   url: "/user_snps",
   data: { snp_name: 'i6000038', local_genotype: '' },
   success: function (html) { $("#user-list").append(html); }
   });
</script>
```

philippbayer:
Oh wow I totally forgot about that! So this issue is moot then :)

gedankenstuecke:
Yeah, only thing that might be of interest would be how to get around the ajax loading when robots hit that page? As I think it loads these things now as well?

philippbayer:
We could use the robots.txt for that? pull out the ajax into a .js file and then  

Disallow: /*.js$
?

gedankenstuecke:
Not sure if that works, but we could try? :D

tsujigiri:
Why not `Disallow: /user_snps`?

gedankenstuecke:
That sounds even easier 😂

PR #419, opened 2017-07-11
Tell bots not to crawl user snps, should ease some load

philippbayer:
As discussed in https://github.com/openSNP/snpr/issues/418

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -3,3 +3,4 @@
 # To ban all spiders from the entire site uncomment the next two lines:
 # User-Agent: *
 # Disallow: /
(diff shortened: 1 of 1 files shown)


gedankenstuecke:
Cool, let's see whether that works :)

tsujigiri:
According to [Wikipedia](https://github.com/openSNP/snpr/commit/7c4d077760b475cea69f95d11b4fcabdef468511#commitcomment-23127947), this disallows anything that starts with `user_snps`, whereas `Disallow: /user_snps/` would be more specific to a directory with exactly that name. Just a tiny detail, but might be cleaner.

PR #420, opened 2017-07-17
Update robots.txt

alphaCTzo7G:


philippbayer:
How come I've never seen this PR?!?!

public/robots.txt
```diff
@@ -3,6 +3,17 @@
...
 # Disallow: /
+
+User-agent: SemrushBot
+Disallow: /
+
+User-agent: AhrefsBot
+Disallow: /
+
+User-agent: *
+Disallow: /snps
```
tsujigiri: Should we keep `/snps` as it contains information people potentially search for?

If anyone has the time, we could try to figure out how much search traffic comes from searches for SNPs. :wink:

philippbayer: Yes, let's keep `/snps`

philippbayer:
Thank you! See #440 for an update, will merge over there

PR #423, opened 2017-08-05
Adds ssl autoupdate scheduler

philippbayer:
This adds a new .sh in the bin folder which checks out the dehydrated repository, requests a new SSL cert from Let's Encrypt and installs it.

TEMPORARILY this is run once per day, tomorrow at the same time we'll know whether it actually works :P


gedankenstuecke:
Let's see what happens tomorrow :D

PR #424, opened 2017-08-07
Fixes bug in autoupdater. Reschedules update to 60 days.

philippbayer:
I am also removing the request for a opensnp.net key since opensnp.net just redirects to opensnp.org


tsujigiri:
While we are at it, should we clone a release instead of master from the dehydrated repo?

philippbayer:
Good idea! Adding now

PR #432, opened 2017-09-05
Should stop bots from using sort and direction sortings

philippbayer:
This disallows from using `sort` and `direction` parameters as bots may cause server outages by asking many times for these computationally intensive views

As in:

```
               "path" : "/snps?direction=desc&page=206694&sort=chromosome",
               "path" : "/snps?direction=desc&page=206694&sort=chromosome",
               "path" : "/snps?direction=desc&page=191983&sort=position",
               "path" : "/snps?direction=asc&page=238042&sort=position",
               "path" : "/snps?direction=asc&page=221971&sort=id",
               "path" : "/snps?direction=desc&page=210945&sort=position",
               "path" : "/snps?direction=asc&page=238042&sort=position",
               "path" : "/snps?direction=asc&page=238042&sort=position",
```

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -4,3 +4,5 @@
 # User-Agent: *
 # Disallow: /
 Disallow: /user_snps
+Disallow: /*sort=
+Disallow: /*direction=


philippbayer:
Let's see how it looks like in an hour.
Do we even need the SNPs view like this? I'm not convinved we need to serve page number 206,694

tsujigiri:
Is there any information on these pages that could help find us via search engines?

philippbayer:
not really! if all SNP's `show` pages are crawled then people googling for SNPs will find those pages more useful/interesting

gedankenstuecke:
Yep, what Philipp says. Though I'm not sure how all `/snp/id` pages would be crawled if the bots don't enter from the main `/snp/` page?

tsujigiri:
But the crawlers find the `show` pages via the `index` pages, right?

gedankenstuecke:
Aye.

tsujigiri:
Maybe some caching could help, instead.

tsujigiri:
Not sure how and when to invalidate cached pages that get each looked at presumably not very often.

tsujigiri:
I think @philippbayer is right. If we disallow sorting it should be alright. Then the database should be able to use the indexes for figuring out the page instead of first sorting by a column we don't have indexes on. On the other hand, people might still want to sort. Maybe we just add the indexes.

tsujigiri:
On the other other hand... who goes through the whole list, other than stupid robots?

philippbayer:
I have a hunch the robots finished going through all the pages and are now going through all permutations of the pages, which would explain why this problem hasn't happened before.

can we make something like a sitemap for the bots so they know where the SNPs' show pages are so the bots can ignore the `/snps` index completely?

tsujigiri:
I don't think they ever finish. It'll just be more bots over time, crawling the pages over and over again.

tsujigiri:
I see these options:
* Add indexes on all columns of the snps table, it can be sorted on in the UI
* Figure out if the request comes from a bot and if so, disable the sorting links or ignore the sorting params

philippbayer:
I think a mix of the two things is good, we really don't need to have the sorting on these giant tables. would be better to keep the current default sort by score like this and disable the rest.

would the indexing on position/chromosome block? do we need it if we disable the columns?

gedankenstuecke:
Yeah, I think the sortings don't make any sense at this point. The default ordering is probably the one that people are interested in, otherwise they'll jump to individual pages right away.

tsujigiri:
Usueally, as a human, you would search anyway, right?

gedankenstuecke:
Yep! :+1:

PR #433, opened 2017-09-05
Remove ability to sort in SNPs index

philippbayer:


philippbayer:
Hmm problematical: you can't press the buttons but the URL parameters like '/snps?direction=asc&page=2&sort=chromosome' are still accepted and work, how do I turn that off?

gedankenstuecke:
You have to remove the parameters from the controllers for that I think :)

app/views/snps/index.html.erb
```diff
@@ -10,16 +10,16 @@
       <p class="text-center snps__text-download">Includes annotation for all SNPs from all sources</p>
     </div>
   </div>
-  <div class="table-responsive">
```
gedankenstuecke: Don't know whether the comment made it: but it should stay responsive as it's about the design for mobiles/tablets

philippbayer:
I 'solved' it in an ugly way now. The responsive part has returned

app/controllers/snps_controller.rb
```diff
@@ -4,6 +4,8 @@ class SnpsController < ApplicationController
   before_filter :find_snp, except: [:index, :json,:json_annotation]
 
   def index
+    params.delete(:sort)
+    params.delete(:direction)
     @snps = Snp.order(sort_column + ' ' + sort_direction)
```
gedankenstuecke: Wouldn't it have been enough to set `Snp.order(ranking: :desk)`, [analogous to this](https://apidock.com/rails/ActiveRecord/QueryMethods/order)? Then the additional parameters wouldn't matter at all?

gedankenstuecke:
Ok, this should make for a nicer fix of the parameters than just deleting them from the param's hash. @tsujigiri I think it's good to go. what do you think?

philippbayer:
Good idea to get rid of the sort_column etc. stuff, it's from *2014*

gedankenstuecke:
Ah, you're still awake, had assumed you went to 💤 ! Looks good to me now.

philippbayer:
Ah et's give it a go, maybe the server will stay up while I sleep :)

app/controllers/snps_controller.rb
```diff
@@ -1,10 +1,11 @@
 # frozen_string_literal: true
 class SnpsController < ApplicationController
-  helper_method :sort_column, :sort_direction
   before_filter :find_snp, except: [:index, :json,:json_annotation]
 
   def index
-    @snps = Snp.order(sort_column + ' ' + sort_direction)
+    params.delete(:sort)
+    params.delete(:direction)
```
tsujigiri: Deleting the params is likely not necessary. By not reading the params, you are already ignoring them.

philippbayer: I thought the same! But playing around with it on my local system the table kept on getting sorted depending on the params in the URL I set manually, I didn't think that was possible

tsujigiri: Me neither... :thinking:

philippbayer: it is a superugly solution and I welcome better ideas :D

gedankenstuecke: Wait, why are these back? With my change to `Snp.order()` this really isn't needed as the parameters are a) never parsed as the helper methods are gone and b) the order command now directly states that the ordering is performed on ranking?

gedankenstuecke: You had to delete them earlier as you didn't change the `order` call, but my commits in this PR changed that ☺️

philippbayer: Hmm i didn't put them back in, must've been github merging commits?

PR #442, opened 2017-11-26
fix Open Humans profile index

gedankenstuecke:
I guess that's on the curse of being too successful 😂 😉 :

We now have over 15 Open Humans profiles linked to openSNP. Which is the point when the pagination kicks in. Only now I figured out that I forgot to add the menu elements that actually allow you to paginate 🤦‍♂️


gedankenstuecke:
You don't have to write any tests if you don't write any code! 🤔

Issue #444, opened 2017-12-06
Zipfulldata.rb crashes


gedankenstuecke:
Related to the crashes mentioned in #443: 

I upgraded our `sidekiq` machine to the next level, which means it now has a nice 8 GB of memory to fullfill its tasks. As a user mentioned via email that the `opensnp_datadump.current.zip` is super outdated I manually started the `Zipfulldata` task and observed the memory consumption. Turns out this is so hungry that even the 8 GB weren't enough, the whole machine crashed once again. 😒 

This might easily explain why also the newsletters can't be sent in a regular interval: The `zipfulldata.rb` is run once a day to create the latest dump. As this just kills the whole machine the newsletters run out of memory at that point. 😂 

I could not fully figure out why the task takes up that much memory. One thing I did notice is that we're heavily over-logging. As every single phenotype/user combination creates a new logging-event. Another thing I noticed: The `create_picture_zip` produces nothing but errors:

```
Dec  5 19:28:55 sidekiq docker/71c7c216ff34[627]: 2017-12-05T18:28:55.153Z 2155 TID-gdihg Zipfulldata JID-5b09eef25e572a91040079c7 INFO: create_picture_zip: Errno::ENOENT: No such file or directory @ rb_file_s_lstat - /home/app/snpr/public/system/user_picture_phenotypes/phenotype_pictures/000/000/057/original/WIN_20141119_184544.JPG
```

I found a very simple reason for this: The whole path `/home/app/snpr/public/system/` is not available in the `sidekiq` Docker container while it is available in the `web` container! Somehow the mounting there doesn't work like it's supposed to.  @tsujigiri @philippbayer Help is appreciated 😱  Do you have an idea for 

1. How to make sure that the `.../system` folder turns up in the `sidekiq` machine? 
2. Should we fix that and remove the super-detailed/senseless logging events and see what happens if we run the thing again?

gedankenstuecke:
What I figured out so far:
1. On the `web` machine we do have a folder `/data2` which is mounted as the `system` directory in the docker container on that machine. 
2. On the `sidekiq` the `/data2` is not mounted at all
3. correspondingly the `system` is never mounted in the `sidekiq` docker container. 😂

tsujigiri:
Maybe the storage box got disconnected again? In that case remounting it should help. Take a look at `/etc/fstab` for the mount-point. I cannot access the machine, currently (memory full again?).

gedankenstuecke:
[...]

PR #445, opened 2017-12-09
Trying to fix the zip

gedankenstuecke:
Trying to just remove the broken pictures for now to see whether this makes the zip creation finally go through. c.f. #444


gedankenstuecke:
It's a super ugly thing, but as the tests would all run through in `dev` it must be somehow related to our deployment. Think we could try merging this, run the `zipfulldata.rb` and see what happens now?

philippbayer:
Yes let's try ^^

philippbayer:
Server went down immediately, but the running Docker image is still 2 days old, looks like coincidence

gedankenstuecke:
Cool, will you start The `sidekiq` task and monitor?

philippbayer:
You mean `rake dump:full`? will do that now

philippbayer:
Hmm looks like it ran out of memory again, server doesn't react right now... have to go to a kid's birthday soon but will keep on monitoring

PR #447, opened 2017-12-10
Put Zipfulldata's file into /tmp first

philippbayer:
~~Cannot merge this yet as not enough space on the `sidekiq` host~~ Host has been resized, 200GB should be enough

While compressing all genotypings the Zipfulldata job the zipfile onto the network storage, and after a while it seems that the network storage hangs and stops replying until you force a reboot.

This PR writes the zip file into local /tmp and then moves the file over to the network folder once zipping has finished, should make it a bit better


gedankenstuecke:
Looks good to me!

PR #448, opened 2017-12-11
Move chmod in zipfulldata

philippbayer:
The chmodding command breaks (access denied?) in the network storage, not 100% sure why.

This moves the chmodding around to an area where I'm sure we're still allowed, then moves it to network storage, mv in ruby preserves the file permissions (on my system).


tsujigiri:
This is likely due to the storage being of the samba kind (windows compatible network shares), not supporting permissions on the level linux does. It's probably also the reason why it's so unreliable... 😒

philippbayer:
Aaah this makes perfect sense, could've thought of this myself - at work some work servers have mounted ~ via Samba, and Perl module installations always die because the tmp folder is in ~/.cpan and most installations try to chmod...

tsujigiri:
We should check whether they added other means of mounting them by now.

PR #450, opened 2017-12-11
Delete the chmod

philippbayer:
I've had it with this broken chmod thing, now it goes


app/workers/zipfulldata.rb
```diff
@@ -56,6 +56,7 @@ def run
       FileUtils.chmod(0o644, @zip_fs_path)
       # move from local storage to network storage
       FileUtils.mv(@zip_fs_path, Rails.root.join("public/data/zip/#{dump_file_name}.zip"))
+
```
tsujigiri: 😂

gedankenstuecke: haha, good catch :D

PR #454, opened 2017-12-12
should fix #416

### LINKED ISSUE #416: Picture phenotype upload broke
Just got a mail, looks like uploading a new picture results in a 404

gedankenstuecke:
This should close #416, the issue was just that we have a form for updating the current picture, the form of course does a `POST` but our route was a `GET`. 😂


gedankenstuecke:
:shipit:

PR #465, opened 2017-12-30
Setup sentry.io for exception tracking

tsujigiri:
I already added the environment variable to the deployment scripts.


strika:
Consider sending notifications to Sentry [asynchronously](https://github.com/getsentry/raven-ruby#async). (If Sidekiq is working reliably.)

Cheers.

tsujigiri:
Interesting. Thanks for pointing that out!

PR #467, opened 2017-12-31
Remove superfluous `respond_to` calls

tsujigiri:
Those are only necessary, if we actually do anything different for the different formats. Rails already looks up the correct templates for the requested formats, if there are any.

This may or may not fix https://sentry.io/opensnp/opensnp/issues/426098731/


philippbayer:
Oh man these must be *years* old

Issue #280, opened 2016-04-28
Ruby version support


tsujigiri:
As it came up on the mailing list: Should we specify a Ruby version to use (using a `.ruby-version` file) or should we try to keep everything compatible with at least 2.0 and up or even 1.9.3 and up? The latter has the advantage that there would be one barrier less to get started. We would add all supported versions to the travis build matrix and (hopefully) be done. 2.3 has a few nice features we wouldn't be able to use, but Ruby works fine without them. ;)

philippbayer:
My Fedora's system Ruby is 2.2.5.
The current Ubuntu long-term support release is "Xenial Xerus" and has Ruby 2.3, with the LTS before that still running 1.9.3  ([source](http://packages.ubuntu.com/search?keywords=ruby])). However, the current LTS was only released a month ago so I'm sure that most places running only LTS releases still run the old one. In that case, 1.9.3 and up would probably be the most "inclusive" choice, what do you think?

gedankenstuecke:
I'm with @philippbayer on that one. I guess we can always drop the 1.9.3 support later on when no one is using it any longer (spoken like a true Python-developer).

philippbayer:
Just ran into this problem during `bundle install`:

```
Installing listen 3.1.5
Gem::InstallError: listen requires Ruby version >= 2.2.3, ~> 2.2.
An error occurred while installing listen (3.1.5), and Bundler cannot continue.
Make sure that `gem install listen -v '3.1.5'` succeeds before bundling.
```

So we can't support older Ruby versions without a bit of gem-circumvention magic (listen comes from Guard - do we need Guard?)

tsujigiri:
Guard is nice for development but not essential. The Guard README claims it is "tested against Ruby 1.9.3, 2.0.0, 2.1.0, JRuby & Rubinius". I suppose they will do something about the `listen` dependency eventually. I'd say remove it for now. I'll try re-adding it when I want to use it. ;)

tsujigiri:
FYI: [This happens if we add 1.9 and 2.3 to the TravisCI config.](https://github.com/openSNP/snpr/commit/9e6a92697506f61fc3ce1d7d58c6bdc4194839f0)

philippbayer:
Everything dies :) At least the 2.3 one seems to be fixable, it just dies when it tries to install 2.3. 

[...]

PR #470, opened 2018-01-01
Remove Ruby version from Gemfile

tsujigiri:
Passenger complained that it couldn't find version 2.3 after I updated
the Docker image to use 2.4 and I'm not sure it makes sense to specify
the version if it has to be a specific one.

Alternatively we could set it to 2.4.3. This seems to even pin it to a patch level automatically. I'd expect this to be more problematic than helpful. What do you think?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,6 +1,5 @@
 # frozen_string_literal: true
 source 'https://rubygems.org'
-ruby '>= 2.3.3'
 
 gem 'dotenv-rails'
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -453,8 +453,5 @@ DEPENDENCIES
   will_paginate
   will_paginate-bootstrap
 
-RUBY VERSION
-   ruby 2.3.3p222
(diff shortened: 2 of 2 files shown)


gedankenstuecke:
We had originally set it in the Gemfile after some discussion in #280 

(btw. the [.travis.yml](https://github.com/openSNP/snpr/blob/master/.travis.yml) still uses `2.3` it seems?)

I guess specifying the minimal version somewhere would be a good idea. But maybe we got to the wrong conclusion back then? I don't have a strong opinion either way. @philippbayer any take on that?

tsujigiri:
Bundler doesn't seem to suppoert a "minimum version" in the first place and complains every time I run it.

gedankenstuecke:
True, so we could remove the 'minimum version'. But then we should maybe provide a `.ruby-version` and update the `.travis.yml` to make sure we test against the version in production? :)

tsujigiri:
Yes, let's do that.

PR #471, opened 2018-01-01
Notify Sentry of new releases

tsujigiri:
Usually Sentry only learns about releases once there is an exception. With these changes, we push this information to Sentry when Rails starts, which lets us keep track of all deployments.


gedankenstuecke:
👍

PR #477, opened 2018-01-07
Get a current dump from production

tsujigiri:
The current SQL dump is causing errors on TravisCI is because it's done with a recent version (by running `db:migrate`), while we use 9.3 on production and TravisCI. I suppose the only way to really fix this is to upgrade Postgres in production, for now this should at least work to fix the tests.


tsujigiri:
Looking for an example, I cannot find them anymore... :ghost:

I would still like to merge this, since I can then set-up Docker Compose with the database version we use in production.

tsujigiri:
Nevermind, it still works even with the errors... 🤷‍

PR #478, opened 2018-01-08
🔥 Remove Rubocop 🔥

tsujigiri:
Since we have Hound, which is more convenient than maintaining the
Rubocop config, let's get rid of it!


tsujigiri:
Hound approves! 🐶

gedankenstuecke:
Travis CI not so much 😉

PR #479, opened 2018-01-09
Trigger Docker build from snpr build

tsujigiri:
In order to build the Docker image on TravisCI instead of Docker Hub, which is faster and has the benefit that the test in the opensnp.org-docker repo ist run on every deploy, we trigger the build over there, which then pushes the image to Docker Hub.

Compare: https://docs.travis-ci.com/user/triggering-builds

I already added the API token to the config.


gedankenstuecke:
👍

PR #480, opened 2018-01-18
Build native recommendify

tsujigiri:
Moved it here from the Dockerfile, so we can simply run it wherever.


tsujigiri:
It's because it depends on this one being merged.

PR #481, opened 2018-01-18
Pass snpr commit hash to docker build

tsujigiri:
When building the docker container, the `git clone` part is cached and not executed again. To actually make it build again, whenever the snpr master changed, we can pass along the commit hash to the other build and use it as an `ARG` in the Dockerfile.

### DIFF
--- a/bin/deploy
+++ b/bin/deploy
@@ -1,9 +1,24 @@
-#!/usr/bin/env sh
+#!/usr/bin/env ruby
 
-curl -s -X POST \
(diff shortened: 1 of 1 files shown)


tsujigiri:
Do not merge, yet, there might be better solutions. Maybe cloning the code outside of docker and using `ADD`? 🤔

tsujigiri:
No, I think that's a good solution. Ready when you are. 😉

PR #488, opened 2018-02-21
Add thanks to Katrin Leinweber @katrinleinweber

philippbayer:
that's all!

### DIFF
--- a/public/humans.txt
+++ b/public/humans.txt
@@ -84,6 +84,7 @@
         Brian Degger
         Malin
         Joerg Rings
+        Katrin Leinweber
 
 /* SITE /*
         Where: https://opensnp.org


tsujigiri:
Our humans.txt doesn't seem like it's properly formatted according to the standard. http://humanstxt.org/Standard.html

philippbayer:
:scream: As in, all people need a 'name:' tag before them?

tsujigiri:
in their own humans.txt, they use the person's title as a key: http://humanstxt.org/humans.txt
The comments are off, too.

tsujigiri:
The "standard" is not very extensive, not sure formatting matters that much. :wink:

philippbayer:
Ah I see, they all need keys then. I don't think we can add titles to /* TEAM */ as everybody does/did a little bit here and there.. Location is also hard because people keep on moving. Just keep it like it is? It would've been good to add a 'why' thank you in the first place as key

tsujigiri:
I actually like that nobody knows anymore who did what. In a way it says that small contributions matter as much as big ones. The standard doesn't even specify whether the keys need to be unique. Even though it looks like it. :man_shrugging:

PR #492, opened 2018-03-27
[WORK IN PROGRESS] - preparsing job recognises uploaded archive better

philippbayer:
- [x] use bash tools to check file type

- [x] use the right gem to extract from the archive

- [ ] make sure everything downstream still works, needs more testing from me

### DIFF
--- a/app/workers/preparsing.rb
+++ b/app/workers/preparsing.rb
@@ -28,22 +28,22 @@ class Preparsing
     when /ASCII text/
       logger.info "File is flat text"
       reader = File.method("open")
-      is_collection = False
+      is_collection = false
     when /gzip compressed data, was/
       reader = Zlib::GzipReader.method("open")
       logger.info "File is gz"
-      is_collection = False
+      is_collection = false
     when /gzip compressed data, last modified/
       reader = lambda { |zipfile| Gem::Package::TarReader.new(Zlib::GzipReader.open(zipfile)) }
-      is_collection = True
+      is_collection = true
     when /POSIX tar archive/
       logger.info "File is tar"
       reader = Gem::Package::TarReader.method("new")
-      is_collection = True
+      is_collection = true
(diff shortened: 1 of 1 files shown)


app/workers/preparsing.rb
```diff
@@ -27,14 +64,13 @@ def perform(genotype_id)
...
-
-    rescue
-      logger.info "nothing to unzip, seems to be a text-file in the first place"
+    else
+      system("cp #{genotype.genotype.path} #{Rails.root}/public/data/#{genotype.fs_filename}")
     end
 
     # now that they are unzipped, check if they're actually proper files
     file_is_ok = false
-    fh = File.open(genotype.genotype.path)
```
philippbayer: How did this ever work?? It parses the original uploaded file the way this looks like, that can be anything before extraction

app/workers/preparsing.rb
```diff
@@ -10,11 +12,40 @@ class Preparsing
...
-    biggest_size = 0
-    begin
-      Zip::File.open(genotype.genotype.path) do |zipfile|
+    logger.info "Starting preparse on #{genotype.genotype.path}"
+    # First, we need to find out which archive or flat text our uploaded file is!
+    # We use the bash tool file for that
+    #
+    # There are two possible outcomes - file is a collection of files (tar, tar.gz, zip)
+    # or file is a single file (ASCII, gz)
+    filetype = `file #{genotype.genotype.path}`
```
raivivek: I saw some `.docx` files as well.. not sure if after extraction or before..

gedankenstuecke: Haha, yeah, some folks aren't great at uploading the right file types. I'd say we should reject anything that's not ASCII post-unzip :)

raivivek: That's true. I also saw a few post-genotype analyses from Promethease.

PR #494, opened 2018-04-03
Use production dump for compatibility with PostgreSQL 9.5

tsujigiri:
The tests in the docker repo failed, because PostgreSQL 9.5 doesn't understand everything that's in the 10.1 dump.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -2,17 +2,16 @@
 -- PostgreSQL database dump
 --
 
--- Dumped from database version 10.1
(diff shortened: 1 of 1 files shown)


tsujigiri:
I'm not quite sure how to handle this in the future. The next migration will yield a Postgres 10 dump again, unless we use 9.5 locally, which isn't trivially achieved. I suppose we should revisit the idea of setting up Docker Compose, which I started but didn't quite followed through.

PR #495, opened 2018-04-08
Remove blocking call from structure.sql

tsujigiri:


philippbayer:
I guess we really don't need this due to this inthe documentation:

```
If untrusted users have access to a database that has not adopted a secure schema usage pattern, begin each session by removing publicly-writable schemas from search_path. For example, add options=-csearch_path= to options, or issue EXEC SQL SELECT pg_catalog.set_config('search_path', '', false); after connecting. This consideration is not specific to ECPG; it applies to every interface for executing arbitrary SQL commands.
```

Since we don't have untrusted users?

tsujigiri:
It doesn't apply to production anyway, since there the migrations are run and that's it. The structure dump is only used in testing/development.

PR #500, opened 2018-05-22
Encrypt message body and subject

tsujigiri:


philippbayer:
Very good idea :+1: - how come te Hound is hanging? Can this be merged?

tsujigiri:
It would be great if someone could double check that the migration does what I think it does. Other than that it should be mergable.

tsujigiri:
Fixed the hound issue. It was still ticked as a requirement, even though the integration changed, which is also why it was in there twice.

tsujigiri:
> key must be 32 bytes

:wink:

philippbayer:
One thing that could be fixed is to add a relevant line to .env.development and .env.test

gedankenstuecke:
That sounds good, it's already the case for `.env.test` but could also be added for `.env.development` 👍

tsujigiri:
Fixed. I also added a key to the deployment script, so this should be ready to be merged.

gedankenstuecke:
Okay, looks ready to be merged to me!

tsujigiri:
I'll just do it then. :wink:

gedankenstuecke:
Let me fix the codeclimate thing :)

gedankenstuecke:
well, looks like that didn't help :D

PR #503, opened 2018-06-14
Remove `set_config` call from structure.sql

tsujigiri:
It's blocking and doesn't really do anything, it seems.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -9,7 +9,6 @@ SET statement_timeout = 0;
 SET lock_timeout = 0;
 SET client_encoding = 'UTF8';
 SET standard_conforming_strings = on;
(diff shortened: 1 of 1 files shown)


philippbayer:
Docs say:

> If untrusted users have access to a database that has not adopted a secure schema usage pattern, begin each session by removing publicly-writable schemas from search_path. For example, add options=-csearch_path= to options, or issue EXEC SQL SELECT pg_catalog.set_config('search_path', '', false); after connecting. This consideration is not specific to ECPG; it applies to every interface for executing arbitrary SQL commands.

So I think too that can go, as we don't have untrusted users

tsujigiri:
It's only for testing, anyway.

Issue #425, opened 2017-08-15
Issues with message system


gedankenstuecke:
From our emails

>Every time I look down the list of messaging I have sent, the page resets to the top of the list every 10 seconds or so.
>
> So if I do not get to a message within the 10 seconds, there is nothing I can do;
I cannot find the message, I cannot highlight it and I cannot see what I wrote.

Will have to verify and investigate.

philippbayer:
bump this one again,  have been reminded that it still exists

tsujigiri:
It only happens if the messages are accessed via the "Messages" link in the user dropdown at the top. If you go to your user profile and click the "Your messages" tab there, it's fine.

philippbayer:
Can anybody still replicate this issue?

I have a *similar* issue, I made a gif, is that the issue? Or is that new? It essentially refreshes the entire page forever if you go too deep into the messages tab (at least I think that's what triggers it). I can reproduce that in a local dev environment, not sure yet what triggers it.

![output](https://user-images.githubusercontent.com/413885/46589863-26ab1a00-cae1-11e8-872f-0d8ebaefc2a3.gif)

PR #512, opened 2018-10-08
Circumvents issue with endless refreshing on #messages tab

philippbayer:
Have a look - does it work on your ends too?

Could fix https://github.com/openSNP/snpr/issues/425

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -9,7 +9,6 @@
 //= require ./jqplot.highlighter.min.js
 //= require ./jqplot.pieRenderer.min.js
 //= require jquery_nested_form
-//= require bootstrap.min
 //= require_self
 // = require_tree .
 
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -9,6 +9,7 @@
 //= require ./jqplot.highlighter.min.js
 //= require ./jqplot.pieRenderer.min.js
 //= require jquery_nested_form
+//= require bootstrap.min
 //= require_self
 // = require_tree .
 
@@ -45,7 +46,6 @@ $(document).ready(function() {
             $('#help_three').hide('slow');
         };
     });
-
     // Jump to tab if link like rs7412080#plos is opened
     $(function(){
         var hash = window.location.hash;
@@ -53,6 +53,6 @@ $(document).ready(function() {
         if (anchor.length > 0){
             anchor.click();
         }
-    });
+});
 
 });
--- a/app/views/layouts/_navbar.html.erb
+++ b/app/views/layouts/_navbar.html.erb
@@ -43,7 +43,7 @@
           <li><%= link_to("Settings", "/users/"+current_user.id.to_s+"/edit", class: "navbar-menu__options", :id => current_user) %></li>
           <li><%= link_to("Edit Your Phenotypes", "/users/"+current_user.id.to_s+"/edit#phenotypes", class: "navbar-menu__options")%></li>
           <li class="divider"></li>
-          <li><%= link_to("Messages ("+current_user.messages.where(:user_has_seen => false).length.to_s+")","/users/" + current_user.id.to_s + "#messages", :method => "get", class: "navbar-menu__options")%></li>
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/application.js
```diff
@@ -9,7 +9,6 @@
 //= require ./jqplot.highlighter.min.js
 //= require ./jqplot.pieRenderer.min.js
 //= require jquery_nested_form
-//= require bootstrap.min
```
tsujigiri: 😮

app/views/layouts/_navbar.html.erb
```diff
@@ -43,7 +43,7 @@
           <li><%= link_to("Settings", "/users/"+current_user.id.to_s+"/edit", class: "navbar-menu__options", :id => current_user) %></li>
           <li><%= link_to("Edit Your Phenotypes", "/users/"+current_user.id.to_s+"/edit#phenotypes", class: "navbar-menu__options")%></li>
           <li class="divider"></li>
-          <li><%= link_to("Messages ("+current_user.messages.where(:user_has_seen => false).length.to_s+")","/users/" + current_user.id.to_s + "#messages", :method => "get", class: "navbar-menu__options")%></li>
+          <li><%= link_to "Messages (#{current_user.messages.where(:user_has_seen => false).length})","/users/#{current_user.id}#messages", class: 'navbar-menu__options', :data => { :reload => true }%></li>
```
tsujigiri: `"/users/#{current_user.id}#messages"` can be written as `user_path(current_user, anchor: 'messages')`. And I suppose the `:data => { :reload => true }` might have something to do with the reloading? :smile:

app/assets/javascripts/application.js
philippbayer: interestingly, removing this causes no problem on my end - I do get an endless 404 when it's kept in

app/views/layouts/_navbar.html.erb
philippbayer: Yeah the :method => "get" causes an endless reload on my end (right now on production too), Firefox 62.0.3 - when I force the reload then the entire page reloads, but only once! and it lands on the correct tab

tsujigiri: Which `:method => "get"`? I don't think it has anything to do with the link. It happens if you put `https://www.opensnp.org/users/15#messages` into your address bar as well. [This](https://github.com/openSNP/snpr/blob/a96b015fcfa1bed6aa23d33739beeceb67d55741/app/assets/javascripts/application.js#L51-L57) looks a bit suspicious.

philippbayer: By method => get I mean the current way it's implement on production - 
That JS snippet is definitely the main 'suspect', getting stuck in an endless loop - let me fix that. My forced reloading seems to not have triggered that JS, thereby circumventing it (on my system?)

tsujigiri: I think what's missing in the JS is comparing `hash` with `anchor` and only redirecting if they don't match.

tsujigiri: Ah, no. Scratch that. It needs to figure out, whether the tab is active.

tsujigiri: What's weird is, that it only does that on the messages tab in the first place. :thinking:

philippbayer: and why does this break only know? I put it in a year ago :P

tsujigiri: I have no idea. 😅

app/assets/javascripts/application.js
philippbayer: I've added that now back in

app/views/layouts/_navbar.html.erb
tsujigiri: This is how to do the tab thing properly, according to the [docs](https://getbootstrap.com/docs/3.3/javascript/#tabs).

```js
$('#myTabs a[href="#profile"]').tab('show')
```

philippbayer: I've put in a new commit which gets rid of the method: get, and now it seems to work, even with the 'old' JS?

philippbayer:
:man_shrugging: computers!

thank you for your help :)

PR #528, opened 2020-05-30
add recaptcha to messages 

gedankenstuecke:
To avoid people automating sending spam to all of our users. didn't adapt tests yet 🙈

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -19,16 +19,15 @@ class MessagesController < ApplicationController
   def create
     @message = Message.new(message_params)
 
-    if verify_recaptcha(model: @user)
+    if not verify_recaptcha(model: @user)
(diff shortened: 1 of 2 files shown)


gedankenstuecke:
Ok, also minimized the complexity and made the hound happy I think. Ready for review & merge.

app/controllers/messages_controller.rb
```diff
@@ -19,6 +19,10 @@ def new
   def create
     @message = Message.new(message_params)
 
+    if !verify_recaptcha(model: @user)
+      (redirect_to "/users/#{current_user.id}#messages") && return
```
tsujigiri, line 23: I'd put the `return` on its own line. Otherwise, it's dependent on what the `redirect_to` returns.

gedankenstuecke: Okay, gave that a go!

PR #533, opened 2020-09-21
Add link to `snps` library

apriha:
Hello! This PR adds a link to the [snps](https://pypi.org/project/snps/) library for parsing SNPs.


gedankenstuecke (approved):

gedankenstuecke:
Thanks @apriha, that's a great addition! We should have added that much earlier, as I remember playing around with `snps` and really liking it to work with the data!

apriha:
Thanks @gedankenstuecke !

PR #541, opened 2022-11-30
Refactor Zipfulldata worker

tsujigiri:
* Breaks up Zipfulldata worker into service classes
* Fixes N+1 queries for phenotype and picture phenotype CSVs
  * moving phenotype CSV generation into database for performance
* Fixes unintentional deletion of unrelated files
* Reduces the time it takes to assemble the zip file from about 10 to about 5 hours, with the bottle-neck being zipping the genotype files


tsujigiri:
@gedankenstuecke @philippbayer Are you aware that, in the phenotypes CSV, users with multiple genotypes have their phenotypes listed once for each of their genotypes? This could skew statistics run over the file, if whoever is running them isn't aware.

gedankenstuecke:
@tsujigiri Yep, but I contains the user ID as well on each row as the primary keys (see https://github.com/openSNP/snpr/blob/f995d368163d3ad3df5b0e20289a8ea30b953a00/app/workers/zipfulldata.rb#L76). 

I think that should make it obvious enough and I think the alternative (of having a genotype-cell which contains multiple entries) might be even uglier for post-processing?

tsujigiri:
@gedankenstuecke It could be separate files, one for the user-specific data and one for the phenotype data and the user ID for referencing. The genotype files have the user IDs in their names already.

tsujigiri:
@gedankenstuecke As a counter argument, somebody out there could rely on the structure and would have to adapt...

philippbayer:
I feel like we can ignore the issue of having multiple genotypes' phenotypes listed several times. I think most people would check for that? I constantly have duplicates in my data. Changing that now would need a special notification

gedankenstuecke:
Yeah, I think ideally we would have gone that way of getting the genotypes out of that file and did that split. Now it might be somewhat annoying for folks like [`snps`](https://pypi.org/project/snps/) and [`GenomePrep`](https://supfam.mrc-lmb.cam.ac.uk/GenomePrep/) that already have pipelines for the existing structure!

tsujigiri:
> I constantly have duplicates in my data.

Now you know why. :smirk:

tsujigiri:
Are the names of the genotype files essential as well? The data can already be found in the phenotype files and I'm not sure file names are such great places to put data.
https://github.com/openSNP/snpr/blob/c40e12350c2248ade546088eeb483b086265cd91/app/workers/zipfulldata.rb#L199-L200

gedankenstuecke:
I fear most people that use the data so far focus on the genotype files and the names to understand them, so I'd be a bit hesitant to change those as well.

tsujigiri:
This should produce the same output as in production now. I specified the order of things queried from the database, which it wasn't before. I compared both versions and they look the same to me. See the tests for defined behavior.

gedankenstuecke (approved):
Thanks, that's really great! I had a look over the tests and it seems like a pretty comprehensive definition of the file, so lgtm! 👍

PR #554, opened 2023-05-19
Disable UpdatePapers worker

tsujigiri:
None of them work and the exceptions deplete the Sentry quota.


gedankenstuecke (approved):
Yeah, I think the API calls might have changed a ton by now for PLOS etc.

PR #557, opened 2023-07-05
Send email even if there was an exception in Preparse

tsujigiri:
When there is an exception in `Preparse`, the user never gets notified. This at least lets them know that something went wrong. I'm still looking for the cause of [this specific error](https://opensnp.sentry.io/issues/4049981093/?project=265211&query=is%3Aunresolved&referrer=issue-stream&stream_index=2).


tsujigiri:
Should we set it to not retry on error? The outcome wouldn't be different the next time around (unless we fix it in the meantime) and with this change, we would send the email on every try... 😬

gedankenstuecke:
Yeah, I think that's a fair point! I think we did do a retry as the task would sometimes unmotivatedly crash in the past. But let's assume that's solved? 😂

